PR #89, opened 2026-06-25
fix: add patch to remove module def

safwansamsudeen:
Migrate used to fail.


suite/patches.txt
```diff
@@ -8,6 +8,8 @@ suite.drive.patches.remove_personal #3
 suite.drive.patches.integrate_with_framework
 # --- mail ---
 execute:frappe.db.delete("Mail Cluster Store")
+# --- writer ---
+suite.writer.patches.remove_frappe_writer_module
```
s-aga-r, line 12:
```suggestion
execute:frappe.db.delete("Module Def", {"name": "Frappe Writer", "app_name": "suite"})
```

PR #108, opened 2026-06-28
fix(mail): sidebar background in mobile view

GursheenK:
frappe-ui `Sidebar` is transparent in dark mode by design. It sits flush on `surface-base` and is set apart only by its border. 

But mail's mobile view re-uses the same `Sidebar` as a fixed overlay over content, so the transparent background lets content show through. Since we add the overlay behavior, this fix forces an opaque `bg-surface-base` only on mobile, leaving desktop untouched. 

If the design intent changes, this can move upstream to `frappe-ui` later. For now, keeping it in our app makes sense.


GursheenK:
@krantheman merging this for now you can take a look and remove later if having mobile-specific classes in frappe ui sidebar makes more sense eventually

PR #113, opened 2026-06-29
feat(sheets): private-by-default with an explicit public link

AsifMulani1:
## Why

A user reported **"Sheets made by me is accessible to everyone."** The ask: follow the access model of docs in Frappe Drive.

Sheets are already private-by-default (`if_owner` + per-user DocShare). The mismatch was the **sharing model**: a single ambiguous **"Accessible to all"** toggle that silently meant *every logged-in user of the site* (`everyone=1` DocShare), with no indication to the owner that a sheet was exposed. Drive's personal docs are **private + named-people sharing + an optional, explicit, view-only public link** — no "all logged-in users" blanket. This PR brings Sheets in line.

## What changed

**Backend**
- New `Sheet.is_public` flag — the single source of truth for public access.
- `get_sheet` is now `allow_guest=True` but reads `is_public` **first**: private sheets keep the exact same permission gate (still 403 for guests/non-owners), and it returns `can_write` so the client can render read-only.
- New `set_sheet_public(name, public)` gated on the **share** right; removed the `everyone=1` branch from `share_sheet` / `unshare_sheet` / `get_sheet_shares`; `list_sheets` returns `is_public`.
- `www/sheets.py` lets a logged-out visitor through **only** for a sheet that's actually public.
- Migration patch (`suite.sheets.patches.v0_1.drop_everyone_sheet_shares`) deletes any pre-existing `everyone=1` Sheet DocShares — revert to restricted rather than silently upgrading to a more-exposing public link.

**Frontend**
- Real **read-only editor mode**: in-cell editing, Delete/clear, fill, cut/paste, mutating shortcuts and autosave are all disabled; guests skip the collab socket and render the static `get_sheet` snapshot.
- **ShareDialog** general access becomes **Restricted / "Anyone with the link"** (view-only), wired to `set_sheet_public`.
- **Transparency**: "Public" / "View only" badges in the editor topbar and a globe indicator on the Home grid + list; Share button hidden for viewers.

## Scope
- Public access is **view-only** — no public edit.
- Live collaboration for anonymous viewers is out of scope (they get a static read-only snapshot).
- Bundle is not rebuilt here (source-only).

## Provenance & validation
- Ported from standalone **frappe/sheets#2** (where the same change passed 59 python + 1041 frontend tests + build).
- In this monorepo I validated syntax only — python `ast.parse`, JS `node --check`, and `@vue/compiler-sfc` compile all pass — because the suite frontend deps aren't wired in my checkout (no `node_modules`) and suite isn't bench-installed locally. **Please let CI run the build + python tests.**
- Backend python tests are included (`test_api_security`, `test_share_notify`). The frontend test port is deferred — suite has no vitest infra for sheets yet.

### DIFF
--- a/frontend/src/apps/sheets/canvas/index.js
+++ b/frontend/src/apps/sheets/canvas/index.js
@@ -8,7 +8,11 @@ import { isWrapText } from '../utils/text-wrap.js'
 
 export { colLabel, cellId, parseCellId } from '../utils/cells.js'
 
-export function createGrid(canvas, { onSelect, onCommit, onInput, onCancel, getFormat, onFill, onBatchCommit, getMergeInfo, isSlave, getMasterId, getComment, getValidation, getCondFormat, getRightInset, onHyperlinkClick, onDropdownClick, onResizeEnd, getSheetNames, getCurrentSheet, getEditingHomeSheet } = {}) {
+export function createGrid(canvas, { onSelect, onCommit, onInput, onCancel, getFormat, onFill, onBatchCommit, getMergeInfo, isSlave, getMasterId, getComment, getValidation, getCondFormat, getRightInset, onHyperlinkClick, onDropdownClick, onResizeEnd, getSheetNames, getCurrentSheet, getEditingHomeSheet, readOnly = false } = {}) {
+  // When true, the grid is a viewer: the in-cell editor never opens, so no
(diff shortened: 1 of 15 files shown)


AsifMulani1:
Closing — Sheets won't ship its own private/public access model. Sharing and visibility settings will be common across the suite via Drive, so the per-sheet public-link feature here is superseded. Not merging.

PR #145, opened 2026-07-03
fix: remove website redirect for root path

diniemasjuki:


suite/hooks.py
```diff
@@ -64,7 +64,7 @@
 
 # mail — website redirects
 website_redirects = [
-	{"source": "/", "target": "/suite"},
+
```
BreadGenie:
```suggestion
```

PR #190, opened 2026-07-10
feat(mail): route accepted senders to Inbox and let spam fall through screening

s-aga-r:
## Why

Reworks step 4 of the automation Sieve script (the screening gate). Previously any sender not on the accepted list was always filed into **Screener** with a `stop;`, so the server's default spam filtering never got a chance to run on unrecognised senders.

## What

The screening gate is now a three-way route:

- **Accepted senders** (and the account's own identity emails, which are always trusted) are delivered straight to the **Inbox**, so accepted mail always reaches the inbox regardless of its spam score.
- Otherwise, mail **not** classified as spam is filed into **Screening**.
- Otherwise (an unrecognised sender whose mail is classified as spam) nothing is done, so the **server's default filtering** assigns the mailbox — e.g. Junk once the spam score exceeds the configured threshold.

## How spam is detected

Spam classification is read with the **`spamtest` extension** (RFC 5235), **not** the `X-Spam-Status` header.

Confirmed against Stalwart's source: in `crates/email/src/sieve/ingest.rs`, Stalwart builds the user's mailbox filter on the **untrusted** runtime and calls `set_spam_status(...)` (Spam/Ham) **before** executing the script — so the verdict is available to the script via `spamtest`. The `X-Spam-Status` header is only stamped afterwards during message ingestion (`crates/email/src/message/ingest.rs`), so it is **not** visible to a user Sieve script (also confirmed by the maintainer in [discussion #2637](https://github.com/stalwartlabs/stalwart/discussions/2637)).

`spamtest` returns a 0-10 value (`Ham -> 1`, `Spam -> 10`; per `stalwartlabs/sieve`), so the gate uses:

```sieve
elsif not spamtest :value "ge" :comparator "i;ascii-numeric" "2" {
  fileinto "Screener";
  stop;
}
```

`:value "ge" "2"` treats Spam (10) and any MaybeSpam (2-9) as spam, and Ham (1) / unchecked (0) as not-spam. `AUTOMATION_SCRIPT_REQUIRE` now includes `spamtest`, `relational`, and `comparator-i;ascii-numeric` (enabled for untrusted scripts by default).

## Notes

- The gate is now an `if`/`elsif` with two `stop;` blocks. `remove_sieve_block` matched only up to the first `stop;`, which would orphan the `elsif` on every rebuild, so it now removes a whole labelled block (from its `# <name>` header to the next header / end of script). Verified it still removes single-action blocks (Reject/Spam) and preserves following blocks.
- Added `get_inbox_mailbox_path` (resolves the Inbox by JMAP `inbox` role, mirroring `get_junk_mailbox_path`).
- Accepted senders now bypass server spam filtering entirely (always Inbox, even if scored as spam).

## Tests

Added `test_remove_sieve_block_removes_multi_stop_block` and `test_remove_sieve_block_preserves_following_block`. All 4 tests in the module pass.

### DIFF
--- a/suite/mail/doctype/sieve_script/sieve_script.py
+++ b/suite/mail/doctype/sieve_script/sieve_script.py
@@ -397,7 +397,9 @@ def has_permission(doc: "Document", ptype: str, user: str | None = None) -> bool
 
 SCREENER_MAILBOX_NAME = "Screener"
 AUTOMATION_SCRIPT_NAME = "frappe_mail_automation"
-AUTOMATION_SCRIPT_REQUIRE = 'require ["fileinto", "imap4flags"];'
+AUTOMATION_SCRIPT_REQUIRE = (
+	'require ["fileinto", "imap4flags", "spamtest", "relational", "comparator-i;ascii-numeric"];'
+)
 
 
 def maybe_build_automation_sieve(account: str, activate: bool = False) -> None:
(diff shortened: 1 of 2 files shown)


s-aga-r:
@greptileai review

PR #209, opened 2026-07-13
fix(writer): show comments at viewport top

### LINKED ISSUE #203: Writer comment at exact top of document will not show up
The check for top is truthy, while it should be `null` if it's detached - as `0` could be a real value.

Dkm0315:
## What changed

- Treat numeric `0` as a valid Writer comment position.
- Reserve `null`/`undefined` for comments that are intentionally unpositioned or hidden.
- Calculate anchor offsets even when an anchor is exactly at the viewport top.
- Advance collision stacking when the preceding card starts at `0`.
- Add helper and mounted Vue regression coverage for visible, stacked, detached, and remote-new comments.

Closes #203.

## Verification

- `yarn test`: 23 files / 357 tests passed
- focused Writer tests: 10 passed (4 helper + 6 component)
- `bench build --app suite`: passed
- `bench --site suite-203.localhost migrate`: passed
- `git diff --check`: passed
- unchanged-upstream browser reproduction: anchored element at `top: 0` produced a hidden card and collapsed sidebar
- fixed browser replay: the same anchor produced a visible card at `top: 42px` and a 288px sidebar
- detached negative: no anchor, no `top` style, hidden pointer/opacity state, and collapsed sidebar

Mutation checks confirmed that restoring the old truthiness condition breaks zero-position visibility/stacking tests, while restoring detached/remote-new `0` sentinels makes a hidden card appear at `200px`.

Before/after screenshots and exact runtime measurements are attached below.

@safwansamsudeen, would you mind reviewing this Writer fix when you have a chance?

### DIFF
--- a/frontend/src/apps/writer/components/FloatingComments.test.ts
+++ b/frontend/src/apps/writer/components/FloatingComments.test.ts
@@ -1,0 +1,237 @@
+import { renderToString } from '@vue/server-renderer'
+import { createApp, h, nextTick } from 'vue'
+import { describe, expect, it, vi } from 'vitest'
+
+vi.mock('frappe-ui', () => {
+  const Stub = { template: '<span><slot /></span>' }
+  return {
+    Avatar: Stub,
+    Button: { template: '<button><slot name="prefix" /><slot /></button>' },
+    Dropdown: Stub,
+    onOutsideClickDirective: {},
+  }
+})
+
+vi.mock('@/boot/session', () => ({
+  useSessionStore: () => ({ user: 'Administrator' }),
+}))
+
+vi.mock('@/apps/writer/composables/useUsers', () => ({
+  useUsers: () => ({
+    getUser: () => ({ full_name: 'Administrator', user_image: null }),
+  }),
+}))
+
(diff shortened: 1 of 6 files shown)


Dkm0315:
## Browser evidence

Reproduced on unchanged `develop` with a real UI-created Yjs-anchored comment, then replayed after the fix on the same document and scroll state.

**Before — unchanged upstream**

- anchor exists: `true`
- anchor viewport top: `0`
- card: `top: 0px; opacity-0 pointer-events-none`
- sidebar width: `0px`

clipboard


**After — fixed build**

- anchor exists: `true`
- anchor viewport top: `0`
- comment text present: `true`
- card: `top: 42px; opacity-100 pointer-events-auto`
- sidebar width: `288px`

clipboard


**Negative check — detached anchor**

After removing the anchor decoration, the card had no `top` style, remained `opacity-0 pointer-events-none`, and the sidebar collapsed to `0px`. This separates the viewport-top fix from intentionally hidden detached comments.

@​safwansamsudeen the browser evidence above is from the isolated fresh bench used for this PR.

safwansamsudeen:
I think this is unnecessarily complicated for a small change.

PR #210, opened 2026-07-14
feat(mail): infinite scroll for All Inboxes

s-aga-r:
Brings the **All Inboxes** view in line with the per-mailbox view (#208): replaces offset pagination with infinite scroll, and adds collapsible date groups.

## Changes

- **Infinite scroll.** Drop the "1–25 of N+" count and prev/next arrows. The accumulated list in the threads resource is the single source of truth: a reset resource replaces it from the top (filter change, initial load); a load-more resource appends the next window (`start = data.length`), deduped by `account + thread_id`. An `IntersectionObserver` sentinel drives loading, and an epoch guard drops stale appends that resolve after a reset.
- **Scroll-preserving refresh.** The Refresh button, the 30s poll, and the `new_mail_created` socket now merge only genuinely-new threads at the top, preserving scroll position instead of jumping to the top.
- **Optimistic row actions.** Archive/trash drop the row optimistically and only refresh counts on success (scroll preserved); a failure resets to the true list.
- **Collapsible date groups**, matching the per-mailbox view. The last group never collapses, since it's where infinite scroll appends new rows.
- **Refresh button moved** to the right of the toolbar, consistent with the other mailboxes.

## Notes

The backend `get_all_inbox_threads` already supports this (`limit`/`start`/`filter_by`), clamping fetch depth to `ALL_INBOX_MAX_FETCH`; the `hasMore && fresh.length > 0` guard stops auto-loading cleanly once the cap is reached or a window adds nothing new.

### DIFF
--- a/frontend/src/apps/mail/pages/AllInboxesView.vue
+++ b/frontend/src/apps/mail/pages/AllInboxesView.vue
@@ -5,7 +5,7 @@
 			<Button v-if="isMobile" icon="menu" variant="ghost" @click="openSidebar" />
 			<Breadcrumbs :items="[{ label: __('All Inboxes'), route: { name: 'mail-all-inboxes' } }]" />
 		</div>
-		<HeaderActions @reload-mails="reload()" />
+		<HeaderActions @reload-mails="refreshThreads()" />
 	</header>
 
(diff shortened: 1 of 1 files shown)


s-aga-r:
@greptileai re-review

PR #236, opened 2026-07-16
fix(calendar): add participants on click, render links and HTML in descriptions

krantheman:
Two fixes in the calendar app, both from feedback on the event popover / modal.

### Selecting a contact didn't add them as a participant

The Participants `Combobox` only listened for `@keyup.enter`, so clicking a contact in the dropdown committed it as the Combobox's own selected value and never reached `addParticipant()`. The address sat in the input, checkmarked but unadded — typing + Enter worked, so the field looked half-functional.

Wired `@update:model-value` to the same `addParticipant()` the Enter path uses, so selections get the same validation, exclusion and duplicate checks. The control then resets so the input clears for the next participant.

### Meeting links and HTML in event descriptions rendered as text

Descriptions used plain `{{ }}` interpolation, so meeting URLs came through as inert text, and invites whose organizer sent HTML showed raw markup (`<p><a href="...">`). Both shapes occur in real invites, so both are handled:

- **Plain text** goes through `<LinkifiedText>`, which splits text into text/link runs and renders each with `{{ }}` — safe by construction, no `v-html`. It already existed under `apps/mail`, so it moved to `src/components` (where calendar and meet already share components) and lost its mail-utils import. Its hardcoded typography moved to the mail call site via class fallthrough; mail renders identically.
- **HTML** is sanitized with DOMPurify and rendered inline.

### Notes for review

- The sanitize allowlist is deliberately tighter than mail's `<EmailContent>`: that renders in an iframe (hence its `<html>`/`<style>` allowances), this renders in our page. `` is excluded too, since remote assets in an invite from an external organizer are a tracking vector — consistent with mail blocking remote images for untrusted senders. Shout if you'd rather keep inline images.
- Links get `target`/`rel` set by walking the parsed DOM after sanitizing, rather than `DOMPurify.addHook`, which is global and would have quietly changed `<EmailContent>` too.
- `line-clamp-3` is less reliable on the HTML branch than on plain text, since sanitized HTML can contain block children. `[&_p]:m-0` keeps paragraph margins from bloating the popover.
- Verified via `yarn build`; not yet driven in a browser against a live HTML invite.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


krantheman:
@greptile review

krantheman:
@greptile review

krantheman:
@greptile review

PR #239, opened 2026-07-16
feat(search): local email-address suggestions (Tantivy)

s-aga-r:
## What

Adds a local, per-account full-text index (Tantivy) that powers **email-address suggestions** — matching the addresses a user has actually corresponded with, sourced from cached mail messages and contact cards, without a JMAP round-trip.

## How

- **`suite/search/base_index.py`** — generic `BaseIndex` over Tantivy: per-`key`/`entity` on-disk index (layout `<base>/<key>/<entity>`), schema-version reconciliation with auto-rebuild, upsert/delete, and two query modes — free-text `search()` and `search_phrase_prefix()` for as-you-type autocomplete.
- **`suite/mail/search.py`** — `EmailAddressIndex`: a shared, cumulative, per-account index keyed by lowercased address (dedup by construction). `search_email_addresses(query)` matches the query as a consecutive, prefix-terminated phrase over each address's name/email, so:
  - `saga` → matches `sagar…` (mid-token prefix)
  - `sagar.s` → matches `sagar.s@…` / `Sagar Sharma`, **not** `sagar@…` or `sagar.v@…`
  - Also `rebuild_email_address_index(account)` — batched rebuild from the caches, runs in a background job by default.
- **JMAP Account** — whitelisted `clear_search_indexes` / `rebuild_search_index` with matching desk actions; indexes are cleared on account delete.
- **Mail Settings** — a "Destroy Search Index" action.
- Indexing is wired into the message/contact cache writes (never blocks caching on an indexing failure).

## Notes

- The index is cumulative (insert/upsert only) — addresses persist as an address book even after a source is evicted from cache.
- Verified end-to-end against a live account (prefix, phrase-prefix, and exclusion cases all behave as expected).
- Also drops a few unused mail deps from `pyproject.toml`.


s-aga-r:
@greptileai re-review

PR #244, opened 2026-07-17
fix(mail): replace whole roles field instead of patching union discriminator

s-aga-r:
## Problem

Assigning or removing a Stalwart account role failed with:

```
error: invalidPatch
  Invalid JSON Pointer path
  Properties: roles/@type
error: update failed
```

`AccountService.add_roles` / `remove_roles` updated roles by patching sub-paths of the `roles` field:

```python
["update", "Account", account, "--field", "roles/@type=Custom", "--field", "roles/roleIds/<id>=true"]
```

The Stalwart CLI turns each `--field key=value` into a JMAP `set` patch where the slash-separated key is a **JSON Pointer path**. But `roles` is a **tagged union** — `@type` is its variant discriminator, not a regular property — and the server rejects patching the discriminator via a pointer path (`/roles/@type`), failing the whole patch.

## Fix

Replace the entire `roles` field with a full JSON object in a single `--field` (the CLI parses a value starting with `{` as JSON), so the union variant is set atomically — the same way `create` builds it. The object is built from the existing `UserRoles` model via `to_dict()`.

This also lets `remove_roles` drop its `roleIds/<id>=false` clearing hack, since replacing the whole object naturally omits removed roles and reverts to `{"@type": "User"}` when none remain.

## Testing

- [x] `python -m ast` syntax check
- [x] Not yet exercised against a live Stalwart server — needs verification by assigning/removing a role on a mail user (path: `suite/mail/events.py → add_account_role → add_roles`)

### DIFF
--- a/suite/mail/stalwart/account.py
+++ b/suite/mail/stalwart/account.py
@@ -371,9 +371,11 @@ class AccountService(StalwartCLI):
 
 		merged = list(dict.fromkeys([*existing, *role_ids]))
 
-		commands = ["update", "Account", account, "--field", f"roles/@type={RoleType.CUSTOM.value}"]
-		for role_id in merged:
-			commands.extend(["--field", f"roles/roleIds/{role_id}=true"])
+		# `roles` is a tagged union; its `@type` discriminator can't be patched via a sub-path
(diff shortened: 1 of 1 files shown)


s-aga-r:
@greptileai review

PR #247, opened 2026-07-17
feat(mail): scope screening backfill to active, configured, fetching users

s-aga-r:
## Summary

Refines the `enable_screening_for_personal_accounts` patch so the backfill targets a precise, deduplicated set of personal accounts instead of every personal account with screening off.

The eligible list is now built from a join across **User Account** → **User Settings** → **User** → **JMAP Account**, keeping only accounts where:

- `JMAP Account.is_personal = 1` and `enable_screening = 0` (as before)
- `User.enabled = 1` — skip disabled users
- `User Settings.username` is set — skip users without JMAP configured
- `User Settings.skip_schedule_fetch_changes = 0` — skip users who opted out of scheduled change fetching

`.distinct()` collapses the per-user User Account rows to one entry per account, since a JMAP Account is shared across every user with access.

Enabling screening relies on the periodic change fetch to keep the Screening folder current, so users who have it disabled (or who are disabled/unconfigured) are intentionally left untouched. The downstream custom-Sieve check, flag set, Sieve rebuild, and per-account error handling are unchanged.

### DIFF
--- a/suite/mail/patches/enable_screening_for_personal_accounts.py
+++ b/suite/mail/patches/enable_screening_for_personal_accounts.py
@@ -1,0 +1,105 @@
+import frappe
+
+from suite.mail.doctype.sieve_script.sieve_script import (
+	AUTOMATION_SCRIPT_NAME,
+	SieveScript,
+	build_automation_sieve,
+	get_active_sieve_script_id,
+)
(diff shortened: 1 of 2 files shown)


s-aga-r:
related to: https://github.com/frappe/suite/pull/245

PR #260, opened 2026-07-19
feat(exchange): rollback imports on failure

### LINKED ISSUE #55: Rollback imports on failure
The rollback mechanism for the JMAP-based Mail Importer (Mail Exchange).
- Create a new Mailbox (name = `MailExchange.name`)
- Upload the mails into the created folder.
- If failure occurs for any reason, delete that folder along with all the imported emails.
- On success, update the `mailboxIDs` as provided in `import_metadata`.

Similar for Calendar Event and Contact Card imports.

s-aga-r:
## Summary

Implements the rollback mechanism from #55 for all three JMAP-based importers (Mail, Calendar, Contacts Exchange).

Each import now runs **staged**: everything lands in one throwaway container first, and is only moved to its real destination once the whole batch has imported successfully. A failure at any point before the move deletes the staging container and everything staged in it — so the account is never left with a half-finished import.

**Per import (identical shape across the three doctypes):**
1. **Create staging container** — a mailbox / calendar / address book named after `Exchange.name`, hidden (`is_subscribed=False`).
2. **Import into staging** — items are created in the staging container only; each item's real destination (`mailboxIds` / `calendarIds` / `addressBookIds`, resolved from metadata, per-item source, or the account default) is remembered in a `created-id → destination` map.
3. **Commit** — move every staged item into its destination in one batched `set` update (staging is replaced by the destination).
4. **Discard** the now-empty staging container on success (a delete failure here is logged, not fatal — the data is already safe).
5. **Rollback** on any failure — delete the staging container with `remove_emails` / `remove_events` / `remove_contents=True`, wiping the container and every item still staged in it.

## Changes

- `jmap/services/calendars/calendar_event.py` — add `set_calendar_ids(mapping)` (per-event `calendarIds`-only patch).
- `jmap/services/contacts/contact_card.py` — add `set_address_book_ids(mapping)` (per-card `addressBookIds`-only patch).
- `mail_exchange.py` — add staging/move/discard/rollback helpers; `_import_batches` now imports into staging and returns the id→destination map (email moves reuse `EmailService.update(..., replace_mailboxes=True)`).
- `calendar_exchange.py` / `contacts_exchange.py` — parallel staging/move/discard/rollback helpers; `_create_events` / `_create_cards` now stage and return `(targets, skipped, failed)`.

Existing idempotency (skip already-present UIDs), batch limits, and the "server rejected everything → fail" guard are preserved.

## Note

JMAP has no cross-object transaction, so the commit is the atomicity boundary: if the *move* itself partially fails, items already moved stay committed while the rest roll back. Full atomicity holds for failures during the staging phase, which is where mid-import failures realistically occur.

Closes #55

### DIFF
--- a/suite/mail/doctype/calendar_exchange/calendar_exchange.py
+++ b/suite/mail/doctype/calendar_exchange/calendar_exchange.py
@@ -363,6 +363,8 @@ class CalendarExchange(Document):
 		os.makedirs(base_dir, exist_ok=True)
 
 		kwargs = {}
+		service = None
+		staging_calendar_id = None
 		try:
 			if self.import_format == "ics" and self.import_file.endswith(".ics"):
 				shutil.copy(import_file, os.path.join(base_dir, os.path.basename(import_file)))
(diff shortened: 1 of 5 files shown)


s-aga-r:
@greptileai re-review

PR #281, opened 2026-07-20
fix(mail): restore login layout and assign roles before User insert

netchampfaris:
Fixes two bugs in the invite-accept flow reached from `/mail/dashboard/members`.

## 1. The invite form is malformed

The invitation email links to `/mail/signup/<request_key>`. Those public routes carry `meta.isLogin`, which in the standalone mail app told `App.vue` to wrap them in `LoginLayout` (logo, centered card, per-route title). The suite port dropped `App.vue`, and e29bb762df then removed `LoginLayout.vue` as dead code — but nothing ever took over the wrapping. `isLogin` was left with no consumer, so every pre-auth view (invite setup, signup, login, forgot/reset password) rendered as a bare, full-bleed form.

- Restore `LoginLayout.vue` and nest the public routes under it.
- Its `title` switch still matched pre-port route names (`Login`, `ForgotPassword`, …), so every page fell through to _"Create a new account"_ — retargeted at the namespaced `mail-*` names.
- The layout renders `<router-view />` as default slot content, so it works both as a route component and, as before, with an explicit slot.

## 2. "No Roles Specified" when clicking Create Account

Suite assigns `Drive User` / `Meet User` in the User `after_insert` hook, which is too late. By then Frappe's `User.validate` has already run [`check_roles_added`](https://github.com/frappe/frappe/blob/develop/frappe/core/doctype/user/user.py), which sees an empty roles table on a `System User` and warns:

> **No Roles Specified** — Newly created user _x@y.com_ has no roles enabled.

An invited non-admin gets no roles from the mail flow itself (`create_user` only passes `["Mail Admin"]` when `is_admin`), so this fired on every invite acceptance. `set_system_user` also runs during validate, so `user_type` was being resolved against roles that were not there yet.

- Move role assignment to `before_insert` via a shared `suite.suite_core.roles.assign_role` helper; Drive Settings and personal-team creation stay in `after_insert`.
- `before_insert` runs ahead of `set_new_name`, so the helper falls back to `user.email` for the reserved-user check.
- The helper dedupes, so an explicitly assigned role is not appended twice.
- Side effect: drops two redundant `User.save` calls per user creation.

## Verification

Browser-verified against a seeded `Mail Account Request` on `suite.localhost`:

- `/mail/signup/<key>` renders the card layout with the invited email prefilled and readonly.
- `/mail/login` shows "Sign in to your account" (previously "Create a new account").
- Authed dashboard routes (`/mail/dashboard/members`) unaffected by the new nested group.
- `create_user(...)` for an invited non-admin: `System User`, roles `['Drive User', 'Meet User']`, **empty message log** (previously the "No Roles Specified" warning).
- Desk-created user with explicit `System Manager` + `Drive User`: no duplicate roles, no warning.
- `yarn vitest run` — 846 tests across 33 files pass.

Note: this bench has Stalwart unconfigured, so the Stalwart-side steps of `MailAccountRequest.create_account` (account + app-password creation) could not be exercised end to end; the Frappe `create_user` step was verified directly.

### DIFF
--- a/frontend/src/apps/mail/components/LoginLayout.vue
+++ b/frontend/src/apps/mail/components/LoginLayout.vue
@@ -1,0 +1,53 @@
+<template>
+	<div class="sm:bg-surface-gray-1 min-h-screen">
+		<div class="relative z-[1] mx-auto py-8 sm:w-max sm:py-32">
+			<div class="flex">
+				<slot name="logo">
+					<div class="mx-auto flex items-center space-x-2">
+						<MailLogo class="inline-block h-7 w-7" />
+						<span class="select-none text-3xl-semibold tracking-tight">
+							Frappe Mail
+						</span>
+					</div>
+				</slot>
+			</div>
+			<div
+				class="bg-surface-base mx-auto w-full px-4 py-8 sm:mt-6 sm:w-[26rem] sm:rounded-lg sm:px-8 sm:shadow-xl"
+			>
+				<div class="mb-6 text-center">
+					<span class="text-center text-xl-medium leading-5 tracking-tight">
+						{{ title }}
+					</span>
+				</div>
+				<slot>
+					<router-view />
+				</slot>
+			</div>
+		</div>
+		<div class="absolute bottom-4 flex w-full justify-center">
+			<FrappeLogo class="h-4" />
+		</div>
+	</div>
+</template>
+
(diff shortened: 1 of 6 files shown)


netchampfaris:
@BreadGenie Not sure why meet tests are failing here.

PR #283, opened 2026-07-20
fix: unify sidebar state and hide storage labels on collapsed

BreadGenie:
also adds collapsible state for calendar


BreadGenie:
@greptileai that's fine since it'll now be the same behaviour as other apps. Also, users can know what's in the sidebar first before collapsing them.

PR #287, opened 2026-07-21
fix(calendar): resolve event masters via baseEventId so fresh events can be edited

krantheman:
Editing or deleting an event immediately after creating it failed until a page reload.

Master enrichment in `get_calendar_events` resolved events through a uid-filtered `CalendarEvent/query`. That filter runs on the server's search index, which is updated asynchronously — a just-created event was returned by the date-range fetch but missed by the uid lookup. The event then reached the frontend without `master_id` (falling back to the synthetic occurrence id, which the server refuses to update and cannot delete: *"Deleting synthetic ids is not yet supported"*) and with `recurrence_rule` as an unparsed JSON string, which failed type validation on update.

Resolve masters through a lightweight get of `baseEventId` instead — the server derives it directly from the id, so it's correct immediately after creation. Round-trip count is unchanged. Each event now also maps to its own master by id rather than by uid, so an organizer copy and an attendee copy sharing a uid can no longer cross-wire.

### DIFF
--- a/suite/calendar/api/__init__.py
+++ b/suite/calendar/api/__init__.py
@@ -6,12 +6,12 @@ from frappe import _
 from suite.calendar.doctype.calendar.calendar import fetch_calendars
 from suite.calendar.doctype.calendar_event.calendar_event import (
 	fetch_calendar_events,
-	get_master_events_by_uids,
 	update_calendar_event,
 )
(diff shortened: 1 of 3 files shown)


s-aga-r:
@greptileai review

PR #316, opened 2026-07-22
feat(mail): keep forwarded emails in the same thread

s-aga-r:
## Summary

Two independent mail changes:

### 1. Keep forwarded emails in the same thread (feature)
Adds a per-account **Keep Forwarded Email In Thread** setting on JMAP Account. When enabled, a forwarded email gets an `In-Reply-To` header pointing to the original email's `Message-ID`, so the mail server assigns it the same thread as the original.

- New `keep_forwarded_email_in_thread` Check field on **JMAP Account** (Outgoing section) + matching controller type hint.
- `MailQueue.validate_forwarded_in_reply_to()` populates `in_reply_to` from the forwarded email's `Message-ID` (fetched via the JMAP email service) when the setting is on and the mail is a forward. It runs after `validate_in_reply_to_id` and leaves `in_reply_to_id` untouched, so the original is marked `$forwarded` (not `$answered`).
- Toggle surfaced in the mail account settings UI (`Account.vue`).

The compose/forward flow needed no change: it already sends `forwarded_from_id` without `in_reply_to`, so the backend picks up the setting automatically.

### 2. Fix blob fetch for Mail Queue (bug fix)
Mail's blob/data stores are namespaced solely by the JMAP account id, and `fetch_blobs` validates the account via `get_user_for_jmap_account`. Two Mail Queue call sites passed a combined `"{user}:{account}"` key, which is not a valid JMAP Account — raising `JMAP account ... does not exist` when reading a queued mail's cached body or MIME message. Now pass the bare `self.account`, matching every other caller.

## Testing
- Python syntax validated; JSON validated.
- Codebase scanned for other combined-key misuses — none found (the two remaining `{user}:{account}` strings are lock names, which are intentional).

### DIFF
--- a/frontend/src/apps/mail/components/Settings/Account.vue
+++ b/frontend/src/apps/mail/components/Settings/Account.vue
@@ -47,6 +47,17 @@
 		>
 			<Switch v-model="destroyNewsletterAfterSubmit" />
 		</SettingsRow>
+		<SettingsRow
+			class="!py-0"
+			:title="__('Keep Forwarded Email In Thread')"
+			:description="
+				__(
+					'Keep forwarded emails in the same thread as the original by referencing it in the In-Reply-To header.',
+				)
+			"
+		>
+			<Switch v-model="keepForwardedEmailInThread" />
+		</SettingsRow>
 
 		<h2 class="text-base-semibold text-ink-gray-8">{{ __('Incoming') }}</h2>
 		<SettingsRow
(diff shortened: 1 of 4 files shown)


s-aga-r:
Enable **Keep Forwarded Emails in Thread** from **Settings → Account**.

This is a per-account setting, so if you have multiple accounts, you'll need to enable it individually for each account where you want this behaviour.

Please note that this setting only affects emails forwarded after it is enabled. Existing forwarded emails will remain in their separate threads.

Screenshot 2026-07-22 at 10 17 33 PM

PR #317, opened 2026-07-23
feat(store): add Search Store support to Store Entry browser

s-aga-r:
## Summary

Extends the read-only **Store Entry** browser to browse Tantivy **search indexes**, alongside the existing Data Store and Blob Store.

A search store is laid out as `<base>/<namespace>/<entity>`, with one Tantivy index per entity and its schema declared in a `SearchStore` subclass. Since the browser is generic and can't know those subclasses, a new schema-agnostic `SearchIndexBrowser` opens each index with the schema persisted on disk (`tantivy.Index.open`), so any index can be listed and read.

## Changes

- **`store/search_store.py`**
  - `list_index_entities(namespace)` — discovers the index directories under a namespace.
  - `SearchIndexBrowser` — opens one on-disk index generically and exposes `count()`, `browse()` (`{key, size}`), and `read(key)` (`{value, size}`). Documents are enumerated via a match-all query; the key is the `id` field (or first stored field) and the value is the stored fields flattened from Tantivy's `{name: [values]}` shape.
- **`suite_core/doctype/store_entry/store_entry.py`** — wires the search store into `load_from_db`, `get_list`, `get_count`, `_entries`, and `_base_path`; adds a store-code map so record names round-trip; and adds `_available_namespaces`, which lists the parent namespaces of index leaves (search nests one extra directory level).
- **`store_entry.json`** — adds `Search Store` to the `store` options and updates the `entity` field description.

## Testing

Verified end-to-end against a real `EmailAddressIndex`: namespace hint, total and entity-filtered counts, list rows, single-document load via the encoded record name, and the missing-document error path all behave correctly.

### DIFF
--- a/suite/store/search_store.py
+++ b/suite/store/search_store.py
@@ -1,4 +1,5 @@
 import hashlib
+import json
 import os
 import shutil
 from dataclasses import dataclass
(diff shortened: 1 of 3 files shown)


s-aga-r:
@greptileai rereview

PR #319, opened 2026-07-23
feat(sheets): homepage filters, recency groups, and server-side pagination

AsifMulani1:
## What

Upgrades the Sheets homepage from a flat, silently-capped listing to a filterable, paginated one:

- **Ownership tabs** — All / My sheets / Shared with me (frappe-ui `TabButtons`), scoped server-side so results stay correct under pagination.
- **Sort control** — Last modified / Name A–Z / Owner (frappe-ui `Select`), whitelisted server-side via an `order_by` → literal map (caller input never reaches SQL).
- **Recency groups** — under the default sort, list rows group into Today / Previous 7 days / Previous 30 days / Earlier using ListView's native grouped-rows mode; collapse state survives Load More appends (`recency-groups.js`, unit-tested).
- **Server-side pagination** — `list_sheets` gains `start`/`limit` (50/page, capped at 100), server-side title search (300 ms debounce), and returns `{sheets, total}` with a permission-aware aggregate count. This removes the old hard `limit=100` that silently hid older sheets.
- **Scroll layout** — in list mode the ListView owns its internal scroll: topbar, filter toolbar, column header, and the `ListFooter` (Load More + "X of Y", page-size picker suppressed) stay fixed while rows scroll. Grid mode keeps body scroll with a centered Load more + count.
- Filtered/tab empty states get dedicated copy; the branded "No sheets yet" block stays reserved for a truly empty account.

## Why

With more than a screenful of sheets the homepage became a single undifferentiated scroll, and past 100 sheets rows silently vanished. Filters + recency landmarks + explicit pagination fix both.

## Notes

- Sole caller of `list_sheets` is Sheets Home; the response-shape change has no other consumers (verified by repo-wide grep).
- The count query uses dict field syntax (`{"COUNT": "*", "as": "total"}`) — required on current Frappe develop, which rejects string SQL functions in SELECT.
- Feature was runtime-verified in the standalone repo against a 106-sheet dataset (pagination, tab counts, grouping, collapse persistence, stale-response guard, SQL-ish search probes); this port is hunk-identical modulo suite's router navigation and API namespace.
- Tests: 15 DB-free python tests for the API window/filter/order_by-injection behavior; 8 vitest cases for the recency bucketing (`.test.ts` so the suite glob picks it up).

### DIFF
--- a/frontend/src/apps/sheets/pages/Home.vue
+++ b/frontend/src/apps/sheets/pages/Home.vue
@@ -67,15 +67,25 @@
       </div>
     </div>
 
-    <!-- Content -->
-    <div class="home-body">
-      <!-- Loading -->
-      <div v-if="loading" class="home-empty">
+    <!-- Filter toolbar — ownership tabs + sort. Hidden on the true-empty
+         state so a brand-new account isn't offered filters over nothing. -->
+    <div v-if="loading || !isTrueEmpty" class="home-toolbar">
+      <div class="home-toolbar-inner">
+        <TabButtons v-model="ownerTab" :buttons="ownerTabs" />
+        <Select v-model="sortBy" :options="sortOptions" aria-label="Sort by" />
+      </div>
+    </div>
+
+    <!-- Loading (initial fetch or a filter/sort/search reset) -->
+    <div v-if="loading" class="home-body">
+      <div class="home-empty">
         <Spinner class="home-spinner" />
       </div>
+    </div>
 
-      <!-- Empty state (no sheets at all) -->
-      <div v-else-if="!sheets.length" class="home-empty">
+    <!-- Empty state (no sheets at all, no filters in play) -->
+    <div v-else-if="isTrueEmpty" class="home-body">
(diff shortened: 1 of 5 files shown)


AsifMulani1:
Pushed `59dd6ac` addressing both review findings:

- **Failed filter resets**: a failed tab/sort/search reset now clears the rows and shows a dedicated "Couldn't load sheets" surface with a Retry button — stale rows from the previous filter can no longer render under the new state. Append (Load More) failures keep the already-loaded rows and surface the error via the topbar badge instead.
- **Timezone-dependent grouping**: `list_sheets` now returns a server-clock `now` in the same naive frame as `modified`, and the recency bucketing compares within that single frame — "Today" is no longer decided by the client's timezone against server-local timestamps.

Both paths verified in the running standalone app (simulated network failure on tab switch → error surface + Retry restores the correct tab; `now` confirmed in the response envelope). Tests updated: the response-shape unit test asserts the `now` field.

PR #323, opened 2026-07-23
feat(sheets): multi-line cells via Cmd+Enter with auto-growing rows

AsifMulani1:
## What

Adds Google-Sheets-style multi-line cells. **Cmd/Ctrl/Alt+Enter** now inserts a newline *inside* the cell instead of committing and moving to the next row, and the row **auto-grows** to fit every line.

Before, the in-cell editor was an `<input>` — physically unable to hold a newline — so Cmd+Enter just committed and moved down, and any pasted multi-line value was clipped to a single line.

## How

- **`canvas/overlay.js`** — the in-cell editor is now a `<textarea>` (`wrap=off`, autosizing), styled identically to the old input so a single line still sits centered and long lines still scroll horizontally. This is what makes real newlines possible.
- **`canvas/index.js`** — the editor's Enter handler checks for a modifier and inserts `\n` at the caret instead of committing. New grid API `autoGrowRowFor(r, val)` grows the row to fit multi-line values (grow-only — never shrinks a row you sized — capped at 400px) and returns the `{before, after}` height diff.
- **`canvas/painters/cell-painter.js`** — committed values containing `\n` render as stacked lines in *every* wrap mode (clip/overflow hard-break on newlines; wrap mode additionally soft-wraps long lines). Blank lines from `\n\n` are preserved.
- **`components/SheetEditor/index.vue`** — on commit, the row-height diff is attached to the existing edit history op (`beforeRowH`/`afterRowH`) so **undo/redo restores value and height as one step**. `_queueOp` destructures only its known keys, so server sync is unaffected; the height persists and version-histories through the existing `viewSnapshot`/`viewRestore` plumbing. A `writeSheet` guard keeps the resize scoped to the current sheet's canvas view.

## Notes

- No new test files (suite's vitest only globs `*.test.ts`; this is behavioral canvas/DOM work with no pure-logic unit to add). Verified end-to-end in the standalone dev stack: newline insertion, per-row auto-grow, multi-line render, and single-step undo/redo of both value and height.
- Consistent with existing manual row-resize behavior in collab (row heights sync via save/reload, not the live op channel) and cross-sheet undo (height is current-sheet view state).

### DIFF
--- a/frontend/src/apps/sheets/canvas/index.js
+++ b/frontend/src/apps/sheets/canvas/index.js
@@ -832,6 +832,20 @@ export function createGrid(canvas, { onSelect, onCommit, onInput, onCancel, getF
     onCommit?.(id, val)
   }
 
+  // A committed value with hard newlines (Cmd+Enter) grows the row so every
+  // line is visible — Google Sheets behavior. Grow-only (never shrinks a row
+  // the user sized). Called by the host on commit; returns the {before,
+  // after} height diff so it can ride the undo op, or null when no growth.
+  function autoGrowRowFor(r, val) {
+    if (typeof val !== 'string' || val.startsWith('=') || !val.includes('\n')) return null
+    const needed = Math.min(400, val.split('\n').length * 16 + 8)
+    const before = rowH[r] ?? DEFAULT_ROW_H
+    if (needed <= before) return null
+    rowH[r] = needed
+    _applyCanvasSize()
+    return { before, after: needed }
+  }
+
   overlay.el.addEventListener('input', () => {
     const val = overlay.getValue()
     onInput?.(cellId(sel.r, sel.c), val)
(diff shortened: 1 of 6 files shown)


AsifMulani1:
Thanks for the review — both issues fixed in the follow-up commit.

**1. Soft-wrapped lines could exceed the row height.** The row sizer counted only hard newlines (`split('\n')`), so a wrap-mode paragraph that soft-wraps to several visual lines was under-sized and clipped. It now counts the *same* visual lines the painter draws.

**2. Fixed 16px pitch overlapped larger fonts.** Both the painter and the row sizer hardcoded a 16px line height; a 24px font overlapped. Line height now scales with the cell's font size.

The root cause of both was two code paths computing wrap/height independently. I extracted `wrapLines` and `lineHeightFor` into `utils/text-wrap.js` as a **single source of truth** — the painter (what it draws) and the row sizer (how tall the row must be) now call the identical functions, so a grown row fits exactly what's rendered, at any font size. `autoFitRow` (double-click row auto-fit) was folded onto the same helpers too, which fixes its own pre-existing hard-newline blind spot and the 16-vs-18px inconsistency.

Added `text-wrap.test.ts` covering `lineHeightFor` (default stays 16px; scales past the font size for large text) and `wrapLines` (hard newlines, blank-line preservation, soft-wrap counting, and the combined case). Verified end-to-end in the dev app: a 24px multi-line cell renders cleanly spaced with the row grown to fit — no overlap.

PR #335, opened 2026-07-23
feat(sheets): Google Sheets keyboard shortcuts + fix whole-line select anchor

AsifMulani1:
## What & why

A real user reported that Sheets was missing common Google-Sheets keyboard shortcuts, and that **Shift+Space felt buggy** because the active cell jumped to column A. This PR closes both gaps.

### New shortcuts (Google Sheets parity)

| Shortcut | Action |
|---|---|
| **Shift+Space** / **Ctrl+Space** / **Ctrl+Shift+Space** | Select row / column / whole sheet |
| **Ctrl+Alt+=** / **Ctrl+Alt+-** | Insert / delete rows or columns from the selection |
| **Ctrl+Shift+1…5** | Number · Time · Date · Currency · Percent |
| **Ctrl+Shift+V** | Paste values only |
| **PageDown / PageUp** | Scroll one screen (Shift extends) |

All are reflected in the in-app shortcut help (`?`).

### Whole-line selection now keeps its anchor (the "buggy" report)

Previously a whole-row/column selection forced the anchor to the top-left corner, so after Shift+Space the active cell dropped to column A and arrows/typing continued from there. Now `selMode` — not the stored corners — drives the perpendicular span, so the highlighted cell **keeps its column/row** and navigation resumes from it, exactly like Google Sheets. `getSelRange()` still reports the full row/column so copy/delete/sum operate on the whole line.

Also: zoom (`Mod+=`) now excludes Alt so `Ctrl+Alt+=` reaches the insert handler; insert/delete match on `e.code` (macOS rewrites `Alt+=`→`≠`); number-format shortcuts match on `e.code` so shifted digits resolve.

## Scope & notes
- **Frontend only**, no backend/schema changes.
- **Ctrl+Shift+6 (exponent)** intentionally omitted — the format engine has no scientific renderer yet.
- Number-format keys reuse the existing `onNumberFormatChange`; insert/delete reuse the existing context-menu handlers (seeded from the live selection); paste-values reuses `doPasteSpecial('values')`.

## Testing
- Full standalone suite green (1304 vitest, incl. 64 canvas tests) — file contents are line-for-line identical to this port.
- Verified live in the running app: row/column/all select (anchor stays in place; ArrowDown after Shift+Space keeps the column; whole-row Count intact), currency/percent formatting, insert/delete column, PageDown, and Ctrl+Shift+V binding.

Ported from the standalone `frappe/sheets` `main` (same change, suite paths).

### DIFF
--- a/frontend/src/apps/sheets/canvas/index.js
+++ b/frontend/src/apps/sheets/canvas/index.js
@@ -165,11 +165,14 @@ export function createGrid(canvas, { onSelect, onCommit, onInput, onCancel, getF
   // ── Selection ────────────────────────────────────────────────────────────────
 
   function getSelRange() {
-    return {
-      r0: Math.min(sel.r, selEnd.r), c0: Math.min(sel.c, selEnd.c),
-      r1: Math.max(sel.r, selEnd.r), c1: Math.max(sel.c, selEnd.c),
-      mode: selMode,
-    }
+    let r0 = Math.min(sel.r, selEnd.r), r1 = Math.max(sel.r, selEnd.r)
+    let c0 = Math.min(sel.c, selEnd.c), c1 = Math.max(sel.c, selEnd.c)
+    // Whole-row / column / sheet selections keep their anchor on the user's
+    // active cell — selMode, not the stored corners, defines the perpendicular
+    // span. So a Shift+Space row still reports every column for copy/delete/etc.
+    if (selMode === 'row' || selMode === 'all') { c0 = 0; c1 = TOTAL_COLS - 1 }
+    if (selMode === 'col' || selMode === 'all') { r0 = 0; r1 = TOTAL_ROWS - 1 }
+    return { r0, c0, r1, c1, mode: selMode }
   }
 
   // Restore a rectangular selection (used by the contextmenu handler when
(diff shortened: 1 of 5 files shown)


BreadGenie (changes requested):
use frappe-ui's shortcut register feature that way you can even show all of these shortcuts easily on a shortcut dialog as well

AsifMulani1:
Added a follow-up commit: **shortcuts now register through frappe-ui's `useShortcut`**, so the keyboard-shortcut help dialog is generated from that one registry (via `getActiveShortcuts`) instead of a hand-maintained list — description lives next to handler, so it can't drift.

- Grid nav/edit shortcuts stay in the canvas, registered as display-only entries so they still appear in the dialog.
- The `e.code` combos (Ctrl+Alt+=/− insert-delete, Ctrl+Shift+1..5 number formats) and the Escape cascade remain in a small residual handler, since frappe-ui matches on `e.key`.
- The dialog reads `getActiveShortcuts()` off the same `frappe-ui` import that registers the shortcuts so both share one registry instance (frappe-ui's own `KeyboardShortcutsModal` imports the registry by relative path → different instance when frappe-ui is pre-bundled → renders empty).
- Verified live on the standalone build; added `useShortcuts.test.ts`.

GursheenK:
> Ported from the standalone `frappe/sheets` `main` (same change, suite paths).

@AsifMulani1 Noticed a few other recent PRs with this description. You can directly open PRs in the suite repo going forward. Would save a lot of chore-like work including building a feature based on the sheets repo paths, then always porting the changes across to suite repo with new paths for the directories. That process is kinda hacky and can totally be avoided since old repos are set to be archived anyway so suite remains the single source for all changes going forward.

BreadGenie (approved):

PR #331, opened 2026-07-23
feat(mail): role-aware routing and setup screens for the Mail app

s-aga-r:
## Why

Opening Mail before setup was complete dropped users on a blank, error-filled page (or bounced them to Desk). This adds friendly, role-aware routing that guides each user to the right place based on their role and how far mail setup has progressed.

## Behavior

**Users without a "Suite User" role** (Administrator / System Manager / Suite Admin — no personal mailbox) → the admin area (`/mail/admin`):
- `server_url` unset → "not configured" screen with a link to Mail Settings (shown to System Managers/Administrator) to configure it.
- `server_url` set but username/password not (`is_stalwart_configured` is `False`) → message that the Admin Dashboard is only accessible after the mail server is fully configured.
- Fully configured (`is_stalwart_configured` is `True`) → Admin Dashboard is visible.

**Users with a "Suite User" role** → `/mail`:
- `server_url` unset → message to ask the administrator to set up Mail.
- `server_url` set but the user's own username/password not set → a page to enter their JMAP account credentials.
- Fully configured (`is_jmap_configured` is `True`) → renders their mail.

## Changes

- `get_user_info` now returns `is_server_configured`, `is_stalwart_configured`, and `is_suite_user`.
- Mail navigation guard split into dashboard (gated on `is_stalwart_configured`) vs. mail-viewing (gated on `is_server_configured` → `is_jmap_configured`) branches, with role-based landing.
- New `NotConfiguredView` (role-based messaging) and `CredentialsSetupView` pages, plus a `/mail/admin` entry.
- New **Credentials** tab in Mail → Settings to manage the JMAP connection (Server URL read-only, Username, App Password, Backup Email — the last moved out of the Account tab).

## Tests

- Frontend route tests extended (not-configured, credentials-setup, admin entry); all mail tests pass.

PR #336, opened 2026-07-23
feat(mail): add Credentials settings tab to manage JMAP connection

s-aga-r:
Add a "Credentials" tab under Mail -> Settings that lets a user view and update their own User Settings connection fields:

- Username and App Password (JMAP connection credentials)
- Backup Email, moved here from the Account -> Recovery section
- Server URL shown as a read-only field (virtual, set by the administrator)

App Password is never prefilled (Password fields come back masked) and is only written on save when the user enters a new value. Saving runs the doctype's JMAP validation, so bad credentials surface inline.

The tab is available to all logged-in users so it can also be used for initial self-setup, and backup_email handling is removed from Account.vue accordingly.

---

Split out from #331 (this was the second of its three commits).

### DIFF
--- a/frontend/src/apps/mail/components/Modals/SettingsModal.vue
+++ b/frontend/src/apps/mail/components/Modals/SettingsModal.vue
@@ -33,6 +33,7 @@ import {
 	Folders,
 	HardDriveDownload,
 	HardDriveUpload,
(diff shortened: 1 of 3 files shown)


s-aga-r:
@greptileai review

PR #345, opened 2026-07-24
test: add e2e tests for drive and writer

BreadGenie:
also move meet e2e to the same dir


safwansamsudeen:
@BreadGenie many thanks!

PR #347, opened 2026-07-24
fix(sheets): gate sheet-tab mutations for view-only users

AsifMulani1:
## Problem

A **view-only** user was shown every sheet-tab mutation as if it were live:

- Right-clicking (or the tab chevron) opened the context menu with **Rename / Duplicate / Protect sheet / Delete**
- Tabs could be **dragged to reorder**
- **Double-clicking** a tab opened the rename dialog
- The **add-sheet `+`** sat there greyed-out with a divider trailing into nothing

None of these persist — autosave and the write endpoints are already read-only gated — so they were dead, misleading affordances. Only the *cell* context menu was suppressed for viewers; the tab bar was missed.

## Fix

Gate every tab write path on write access, mirroring the existing `_onCanvasContextMenu` guard. A viewer can switch tabs and nothing else (matching Google Sheets):

- `_onTabMenu` wrapper suppresses the tab menu for both openers (right-click + chevron)
- chevron hidden via `v-if` so there's no dead dropdown button
- `:draggable="!readOnly"` disables tab reorder
- `openRenameDialog` early-returns for viewers (covers the double-click path, independent of the menu)
- add-sheet button **and its divider** hidden for viewers rather than rendered disabled

## Visual polish

With the chevron gone for viewers, a `.sn-tab--static` modifier restores symmetric label padding (the tight `2px` right padding only makes sense butted against a chevron) and swaps the `grab` cursor for `pointer` (viewer tabs switch on click but can't be dragged).

|  | Editor | Viewer |
|---|---|---|
| Tabs visible / switchable | ✓ | ✓ |
| Drag-reorder | on | off |
| Chevron / tab menu | present | hidden / suppressed |
| Add `+` | present | hidden |
| Label padding | hugs chevron | balanced |

Frontend-only, single file. Verified in the built bundle by toggling `get_sheet` `can_write`: the editor keeps all four actions + drag; the viewer shows tab pills only — no menu, no chevron, `draggable=false`, double-click opens no rename dialog — and the bar is balanced and centered.

### DIFF
--- a/frontend/src/apps/sheets/components/SheetEditor/index.vue
+++ b/frontend/src/apps/sheets/components/SheetEditor/index.vue
@@ -2593,18 +2593,25 @@ function onTabClick(name) {
 const tabDragName = ref(null)
 const tabDragOver = ref(null)
 
+// Reorder is a mutation. `:draggable="!readOnly"` stops a viewer starting a
+// drag, but dragover/drop still fire when *another* element is dragged onto a
+// tab — guard the whole lifecycle so the invariant "viewers can't reorder" is
+// explicit here, not just an emergent effect of the null tabDragName.
 function onTabDragStart(e, name) {
+  if (readOnly.value) return
   tabDragName.value = name
   e.dataTransfer.effectAllowed = 'move'
   // Some browsers require setData to allow the drag
   try { e.dataTransfer.setData('text/plain', name) } catch (_) {}
 }
 function onTabDragOver(e, name) {
+  if (readOnly.value) return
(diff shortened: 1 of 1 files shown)


AsifMulani1:
Addressed the drag-handler finding in cef1c37f2: added an explicit `readOnly` early-return to `onTabDragStart`/`onTabDragOver`/`onTabDrop`. They were already inert for viewers (they key off `tabDragName`, which only a draggable tab sets, and the server rejects any reorder), but the guard was emergent rather than stated — now it's explicit at each entry point.

PR #355, opened 2026-07-27
feat(sheets): surface sheets in Drive as content files

AsifMulani1:
## Why

Sheets is currently a fully parallel app: a `Sheet` has no Drive presence, so it never shows up in — or opens from — Drive, and Drive's **New** menu has no way to create one. This makes Sheet a first-class **Drive content-app**, using the same `content_doctype` / `content_docname` mechanism Writer and Slides already use.

## What

**Backend**
- `Sheet.after_insert` backs every sheet with a Drive `File` (mime `frappe/sheet`, file_type `Spreadsheet`) via `File.create_for_doc`. A new `create_sheet(parent)` endpoint lets Drive create one directly in the folder the user is looking at. `title_field` is set on the doctype so the File inherits the sheet's title rather than its hash name.
- **Lifecycle sync** so the backing File never dangles:
  - rename (`rename_sheet`) **and** in-editor rename (autosave writes the title via `db.set_value`, which skips `on_update`) keep the File's name in step through `sync_content_file_title` — deduped within the folder, so a duplicate title **renames to `Title (1)` instead of throwing**.
  - soft-trash / restore flip the File's status (`set_content_file_trashed`);
  - hard-delete removes it via the existing `on_trash` doc-event.
- `integrate_with_drive` patch backs **existing** sheets (idempotent).
- Two small additions to the content-app SDK in `drive/overrides/file.py` (`sync_content_file_title`, `set_content_file_trashed`) that any status-trash / db.set_value-titled content app can reuse.

**Frontend**
- **New › Spreadsheet** (gated on the sheets app), plus open + share-link routing to `/sheets/<id>` (keyed on `content_doctype === 'Sheet'`, since `file_type: 'Spreadsheet'` is shared with uploaded `.xlsx`).
- Native sheets render the **Frappe Sheets brand logo** in Drive rows/grid/search, distinct from uploaded `.xlsx/.csv` which keep the generic spreadsheet icon.

## Deliberate scope (follow-ups)

- **Sharing stays on the sheet's own model** for now — the Drive File is a listing/opening pointer, not the permission source. Unifying Sheets onto Drive permissions needs a share-migration and is intentionally out of this PR.
- Drive → Sheet trash is one-way (trashing the *sheet* trashes its File; trashing the File from Drive doesn't trash the sheet).

## Verification

Driven end-to-end on a local suite bench (develop tip):
- New › Spreadsheet creates + opens the editor; the sheet appears in Drive Home with the brand logo and a correct `/sheets/<id>` link; clicking it reopens the editor. 0 console errors.
- rename (editor-save + `rename_sheet`), duplicate-title **dedupe** (`Title (1)`, no throw), soft-trash → File `Trashed`, restore → `Active`, backfill patch (idempotent) — all confirmed against the DB.
- All Sheets tests green (list/permissions/api-security/collab/retention/boot/www/share-notify); Slides `test_presentation` + `test_file` and Drive storage tests green (content-app SDK unaffected).

### DIFF
--- a/frontend/src/apps/drive/components/ActivityTreeItem.vue
+++ b/frontend/src/apps/drive/components/ActivityTreeItem.vue
@@ -28,7 +28,7 @@
       v-else-if="!strikeThrough"
       loading="lazy"
       class="h-full"
-      :src="getIconUrl(entity.file_type)"
+      :src="getEntityIcon(entity)"
       :draggable="false"
     />
     <span
@@ -39,7 +39,7 @@
   </div>
 </template>
 <script setup>
-import { getIconUrl } from '@/apps/drive/utils/files'
+import { getEntityIcon } from '@/apps/drive/utils/files'
 
 defineProps({
   title: {
--- a/frontend/src/apps/drive/components/GridItem.vue
+++ b/frontend/src/apps/drive/components/GridItem.vue
@@ -35,7 +35,7 @@
           v-if="showTypeIcon"
           loading="lazy"
           class="h-4 w-auto"
-          :src="getIconUrl(file.file_type)"
+          :src="getEntityIcon(file)"
           :draggable="false"
         />
         <p class="truncate">
@@ -51,7 +51,7 @@
   </div>
 </template>
 <script setup>
-import { getIconUrl, getThumbnailUrl } from '@/apps/drive/utils/files'
+import { getEntityIcon, getThumbnailUrl } from '@/apps/drive/utils/files'
 import { ref, computed } from 'vue'
 const props = defineProps({ file: Object })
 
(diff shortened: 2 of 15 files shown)


AsifMulani1:
Fixed both filename-collision paths in `21dbbe6`:

- **Creation bypassed folder-level dedup** → `File.create_for_doc` now runs `get_new_file_name(title, folder, file_type)` before creating, so like-titled docs get `Title` / `Title (1)` instead of two identical Drive entries. This is in the shared content-app SDK, so `Sheet.after_insert`, `create_sheet`, and Slides/Writer all get it consistently. `get_new_file_name` only counts **active** siblings, so it's a no-op when the name is free.
- **Restore could reactivate a taken name** → `set_content_file_trashed` now dedupes on restore before the file re-enters the active namespace (the trashed file isn't counted, so a free name is preserved).
- **Backfill patch** dedupes too, so importing many "Untitled Spreadsheet" sheets doesn't produce collisions.

Verified live: two sheets titled "Dup Test" → `Dup Test` + `Dup Test (1)`; trashing one, creating a third that takes the freed name, then restoring → the restored file dedupes to `Dup Test (2)` rather than colliding. All Sheets + Slides (`test_presentation`, `test_file`) tests stay green.

`sheet.py` itself needed no change — its `after_insert` delegates to `create_for_doc`, which is where the dedup now lives.

Issue #136, opened 2026-07-01
feat: improve admin dashboard


s-aga-r:
++ User/Account
++ Mailing List
++ Group
++ Role (?)

PR #357, opened 2026-07-28
feat(mail): build out the Mail Admin Dashboard

s-aga-r:
Builds out the Mail Admin Dashboard: detail pages for the directory resources, and the Emails, Reports, Observability and Actions sections.

## Directory

- Member, group, mailing list, role, domain, DKIM signature and OAuth client detail pages, each editable in place.
- Members support multiple addresses, per-address full names and enabled toggles, their own quota action, and password/status actions. Groups and mailing lists mirror that layout; members, groups and mailing lists cross-link to one another.
- The Stalwart directory is now managed over JMAP rather than by shelling out to the CLI.

## Emails, Reports, Observability, Actions

- Queue: a queued message page with fully editable recipients.
- Reports: inbound/outbound DMARC, TLS and ARF listings with detail pages.
- Logs: searchable list with per-entry detail.
- Actions: the server's management operations, grouped by section, two per row, with an input dialog for the parameterised ones.

## Behaviour worth calling out

Several fixes came from checking what the server actually accepts, rather than from the UI alone:

- **Report filtering** matched no property the server supports for received reports, which failed with `unsupportedFilter`. Only DMARC is searchable — free-text on inbound, exact domain on outbound — so search is offered there and nowhere else.
- **Log pagination** past the first page failed with `"Pagination is only possible using anchors for logs"`; the log store rejects offset paging when a total is requested and silently repeats page one when it is not. Logs now page by cursor, and services can declare themselves cursor-paginated.
- **DMARC troubleshooting** sent an `RCPT TO` field that is not a property of the action, so filling it in failed the whole call with `Invalid property`. The inputs now mirror the server's own form.
- **Spam classification** gained the remaining inputs. Recipients are a JMAP set on the wire, so they are entered as rows and encoded when the action runs; the MAIL FROM parameters come from the schema, because one variant folds its description into the name and that whole string is the only accepted value.
- **Pausing the MTA queue** stops delivery server-wide, so it is restricted to the Administrator and enforced in `run_action`, not just hidden in the UI.
- **Labels come from the server schema** — log levels and events, action names, enum choices — so they stay accurate across server versions instead of being restated in the frontend.
- **`policyIdentifier`** is a u64 that overflows JavaScript's safe integer range and is sent as a string; left as a number the browser rounded it.

## Known gaps

- The log `text` filter does not appear to narrow results, and the reported total ignores it. Not addressed here.
- `createResource` returns a cached resource and discards the new options, so a revisited list keeps the unmounted component's closures. The logs list opts out; the other listings are unchanged.


s-aga-r:
Part 1 of https://github.com/frappe/suite/issues/136

s-aga-r:
@greptileai review

s-aga-r:
Screenshot 2026-07-28 at 5 47 54 PM

PR #359, opened 2026-07-28
fix(drive): resumable downloads + async folder-zip build

safwansamsudeen:
Serve download bytes from storage (S3 presigned URL / nginx / disk) with HTTP Range support instead of streaming through a gunicorn worker.

- Fixes large folder zips truncating into a corrupt archive when the transfer outran the worker timeout.
- Fixes large single files restarting from zero on a dropped connection.

Folders build in a background job to a temp artifact; the client polls a capability token and downloads the finished, Content-Length'd archive. Artifacts expire hourly. Local-disk deploys can opt into nginx `X-Accel-Redirect` via `drive_xaccel_prefix`; otherwise `send_file` streams off disk with Range support.


safwansamsudeen:
@greptileai can you review this PR?

PR #361, opened 2026-07-28
feat(drive): remove the team concept

safwansamsudeen:
Removes the team concept from Drive: one flat site tree instead of per-team workspaces.

## Model

**Default-deny.** Nothing is readable without a grant. Read is not inherited from the site root — a `Site` folder under it carries the `$GENERAL` read grant and holds shared content, so user folders sitting beside it are private by construction and need no deny rows.

**Resolution is two rules.** Nearest folder wins; within a folder the most specific principal wins:

```
user  >  $GROUP:<name>  >  $GENERAL  >  ""(link)
```

A grant flows down to every descendant until an explicit deny interrupts it. Only the group tier can hold several rows at one node, so ordering deny-first there makes a deny beat a grant among groups without affecting the other tiers.

**Ownership is a row.** Creating a user folder writes an owner grant on it, so ownership inherits like any other permission instead of being a special case in the resolver — and a file someone else puts in your folder is still yours to manage.

**Sharing with user groups.** `Drive Permission.user` accepts a `$GROUP:<name>` sentinel resolved against frappe's `User Group`. The caller's principals are resolved once per call, so the recursive walk gains no join. Groups appear in the share dialog beside people, labelled by id with their member count.

**Every File is Drive-managed.** `is_site_file` and the dual code paths are gone. Attachments are the one carve-out: they resolve explicit Drive rows, else delegate to their reference document, preserving the framework contract content apps rely on.

**Quota is per-user** — site default on Drive Disk Settings, per-user override on Drive Settings, 0 = unlimited.

**Routing** is file/folder-scoped only (`/f/`, `/d/`, `/g/`); `/t/:team` redirects home. Teams UI is replaced by site-user management with admin-gated invites.

## Migration

`remove_teams` (idempotent, runs while the team tables still exist): personal team roots become user folders, every other team moves into a `Previous Teams` folder with membership rewritten as permission rows for an admin to reorganise. Only a public team gets an explicit `$GENERAL` read. Members reach their old team through **Shared with me**.

`drop_team_doctypes` then removes the doctypes and every leftover team column — frappe never drops columns or custom fields when they disappear from source. `dedupe_drive_permissions` collapses duplicate rows and adds the unique `(entity, user)` index the tier ordering assumes.

Blobs are not moved. `file_url`s are absolute and keep resolving, and a new file inside a migrated folder derives its path from that folder's stored `file_url`, so storage stays internally consistent. The cost is cosmetic: directories keep their old team names, so the disk layout no longer mirrors the logical tree.

Relocating them means moving every blob and rewriting every `file_url`. That is data movement rather than a metadata change — it cannot be re-run cleanly if it fails partway, so it belongs in its own PR with its own rollback story.

## Verified

Migration exercised end-to-end on a site built from scratch on `develop` with real team data, then migrated on this branch. Permission ladder covered: own folder, cross-user denial, inherited grants, access levels, public team, guest denial, group grants and the group deny tie-break, plus the sharing and invite surfaces. Drive e2e 41/41, frontend unit 1079/1079.

## Notes for review

- Sites that ran an earlier build of this branch are stranded (patches logged, team doctypes gone, no `Site` folder) and need a rebuild.
- Search is tree-wide and resolves permissions per row; recall and paging are product decisions left as-is.
- "Restricted" on a child revokes per-user grants inherited from a parent. Correct per nearest-wins, worth knowing.


safwansamsudeen:
@greptile-apps could you review the PR in full?

safwansamsudeen:
@greptileai could you review, bypassing the 100 file limit?

PR #378, opened 2026-07-29
feat(drive): list view on frappe-ui/list, inline folder tree, derived breadcrumbs

safwansamsudeen:
## Summary

Drive's list view now runs on `frappe-ui/list`, folders expand inline as a tree, and breadcrumbs are derived from the route instead of a mutable global. Plus Writer list-numbering and tab fixes.

## Changes

**List view on `frappe-ui/list`** — real column tracks, inset dividers, a sticky sortable header, and the typography/dialog cleanups that came with it.

**Folders expand inline.** Hovering a row swaps the folder icon for a chevron; clicking it expands the folder underneath. Clicking elsewhere in the row still navigates in. Nested rows behave like any other row — selection, context menu, rename, move, delete, drag-and-drop in and out of a subtree.

**Breadcrumbs are derived.** The trail was a mutable global written from eight places, and it doubled as ambient state — the sidebar highlight and the `personal` flag on upload both read crumb 0, so a missed writer left it stale or empty. It is now computed from the route plus the entity the open page publishes. This fixes the navbar vanishing mid-navigation, and moving a folder updating its trail (which never worked; a loading placeholder hid it).

**Writer** — list numbering no longer restarts when you edit around a list, and reordering tabs can't duplicate them.

## Notable fixes along the way

- Rename surfaced an empty error toast and left the new name on screen after a failure.
- `MoveDialog` closed before awaiting the move, so its `@complete` refetch never ran.
- Presentations crashed the list view: the API returns `title`, not `file_name`.
- Inline rename was unreadable in dark mode (`bg-surface-white` is a retired token, so the background was dropped).
- Keyboard shortcuts lived in two divergent copies, one per view.

## Testing

- `folderTree.test.ts` — 16 cases covering fetch params, the stale-response race, collapse-drops-subtree, path keys, skeleton counts.
- `breadcrumbs.spec.ts` — 7 e2e cases: section names, nested trails, ancestor navigation, deep links, browser back, rename, move.
- `rename.spec.ts`, `tabs.spec.ts` — breadcrumb rename persistence; tab reordering preserves content.

Full suite: 1057 unit tests, 34 e2e.

## Known limits

⌘A and the filter chips apply to top-level rows only, socket updates patch just the current folder, and the chevron isn't keyboard-reachable yet. The navbar still flashes on navigation because `DriveLayout` keys its `router-view` on `$route.fullPath`, remounting the page and navbar on every route change — the skeleton makes it read as loading, but the remount itself is untouched.


safwansamsudeen:
Also redoes list view, adds infinite scrolling, removes rename dialog, and a couple more long pending issues.

PR #382, opened 2026-07-29
fix(mail): show an error page when the mail server is unavailable

### LINKED ISSUE #43: Error page for when mail server is down
Instead of white screen of death

s-aga-r:
Closes https://github.com/frappe/suite/issues/43

When Stalwart is temporarily down, every request failed with an opaque 500 that only surfaced in the browser console — and on initial load the mail app rendered a blank page, because the router guard's awaited mailbox fetch rejected and aborted the first navigation before the app ever mounted.

## Backend

- `JMAPConnection` (the single choke point for all Stalwart traffic — user JMAP and admin/management) now maps connection failures (connection refused, DNS, timeouts) and reverse-proxy gateway responses (502/503/504) to a new `MailServerUnavailableError`.
- The error carries `http_status_code = 503`, so Frappe responds with `503 Service Unavailable` + `exc_type` instead of a generic 500, making outages distinguishable from application bugs. The original `requests` exception is chained for diagnosis.
- Ordinary error statuses (4xx, plain 500 from Stalwart) still raise `HTTPError` unchanged.

## Frontend

- The suite-wide `resourceFetcher` (`boot/config.ts`) detects the `exc_type`, rewrites the error to a friendly translated message (so existing `onError` handlers that toast `error.message` / `error.messages[0]` show clean text), and sets a reactive `mailServerUnavailable` flag.
- Since nothing in mail works without the mail server, `MailLayout` swaps the entire route group (including `noLayout` pages) for a dedicated **"Mail server unavailable"** page with a Retry button, instead of decorating a UI whose every fetch would fail.
- Surfaces without a full-page state (e.g. calendar) get one deduped toast on first detection.
- The mail router guard now tolerates a rejected mailbox fetch, so the initial navigation completes and the app mounts to show the error page (previously: blank page). With no mailbox list to validate against, the requested route is kept rather than redirecting.

## Test plan

- Stop Stalwart, load `/mail` → the unavailable page renders (no blank page); Retry reloads.
- Stop Stalwart mid-session, perform any action → page swaps to the unavailable state.
- 400s from Stalwart still surface as before; `tsc`, the mail vitest suite (24 tests) and the production build pass.

### DIFF
--- a/frontend/src/apps/mail/components/MailServerUnavailableView.vue
+++ b/frontend/src/apps/mail/components/MailServerUnavailableView.vue
@@ -1,0 +1,24 @@
+<template>
+	<div
+		role="alert"
+		class="flex h-dvh flex-col items-center justify-center gap-4 bg-surface-base p-8 text-center"
+	>
+		<CloudOff class="h-10 w-10 text-ink-gray-5" />
+		<div class="flex flex-col gap-1.5">
+			<h1 class="text-2xl font-semibold text-ink-gray-9">{{ __('Mail server unavailable') }}</h1>
+			<p class="text-base text-ink-gray-6">
+				{{ __('The mail server is temporarily unavailable. Please try again in a few minutes.') }}
+			</p>
+		</div>
+		<Button variant="solid" @click="reloadPage">{{ __('Retry') }}</Button>
+	</div>
+</template>
+
+<script setup lang="ts">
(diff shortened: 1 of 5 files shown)


s-aga-r:
Screenshot 2026-07-29 at 2 27 16 PM

PR #390, opened 2026-07-29
fix(mail): keep the composed mail across the mobile breakpoint

### LINKED ISSUE #38: [mobile] Mail disappears when resized between breakpoints
I was resizing the window due to another bug (enter wasn't working for some reason) and I lost my entire mail 😭. I suspect you use different elements for mobile and desktop layout?

edit: turns out it only disappears from the UI, so reloading the page got my email back.

krantheman:
Crossing 640px while composing swapped the composer's wrapper (`Dialog` ⇄ mobile sheet), unmounting the editor and with it the `mail` reactive holding everything typed — the replacement instance opened blank (the text survived only as an auto-saved draft, hence the reload trick).

The draft now lives in `SendMail`, above the swap, and is handed to whichever editor instance takes over; it's released when the composer closes so the next one opens fresh.

Closes #38

### DIFF
--- a/frontend/src/apps/mail/components/ComposeMailEditor.vue
+++ b/frontend/src/apps/mail/components/ComposeMailEditor.vue
@@ -307,12 +307,6 @@ const insertContacts = (insertInto: string) => {
 	showContactsModal.value = true
 }
 
-const showCcBcc = ref(!!mailDetails?.cc?.length || !!mailDetails?.bcc?.length)
-const toggleCcBcc = () => {
-	showCcBcc.value = !showCcBcc.value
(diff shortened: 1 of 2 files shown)


krantheman:
Superseded by #394.

This only fixed desktop → mobile. The draft ref lives in `SendMail`, which survives that direction because `HeaderActions` keeps it mounted outside its `v-if="!isMobile"`. Going the other way, the composer belongs to `MobileTabBar` — itself `v-if`'d on `isMobile` in `DefaultLayout` — so the whole `SendMail` is destroyed and the draft with it.

#394 stops the tree swapping at all rather than trying to carry state across it.

PR #400, opened 2026-07-29
feat(mail): actually delete Trash and Junk after 30 days

krantheman:
Trash and Junk show "Items in this mailbox will be automatically deleted after 30 days", but nothing implemented it — the only way mail left those mailboxes was the **Delete Now** button beside the banner. This adds the daily job that makes the banner true.

### The clock

JMAP has no "moved to mailbox at" timestamp — an Email only carries `receivedAt`. Purging on that would erase a two-year-old mail the night after a user trashed it, with no window to restore it. So the job keeps the clock itself: it records when it first *observed* each message in Trash/Junk (in the account's data store) and deletes once that stamp is 30 days old.

Deriving it from mailbox contents rather than from intercepting moves makes it self-healing — works whether the mail was trashed from our UI, over IMAP, or by another JMAP client, and a restore-then-retrash starts a fresh window because the stamp is pruned while the message is away.

Stamps for purged ids are deliberately left for the next run to prune as stale, so a delete that failed server-side is retried on its original stamp instead of being handed another full window.

### Notes

- First run after deploy (and after a store wipe) stamps everything already in Trash, so nothing is purged for 30 days. For a job that deletes mail, erring toward keeping it too long is the right direction.
- Retention is overridable per-site via `mail_trash_retention_days` in site_config, floored at one day so a misconfigured `0` can't turn Trash into an instant hard delete. The banner text is not parameterised — the knob is an ops escape hatch, not a user setting.
- Per-account failures are isolated and logged; accounts are fanned out to the long queue in batches of 25.

### Testing

19 unit tests over the retention matrix (`suite/mail/tests/test_retention.py`) — expiry, boundary, restore-then-retrash, unreadable stamps, pruning, both mailboxes, config floor.

Smoke-tested against a live site with 4 JMAP accounts: first run stamped 2,245 messages and purged none; a second run the same day was a no-op; a run with the clock advanced 40 days handed every stamped message to the delete path.

No UI changes.

### DIFF
--- a/suite/hooks.py
+++ b/suite/hooks.py
@@ -265,6 +265,7 @@ scheduler_events = {
 		"suite.sheets.versioning.tasks.truncate_op_log",
 		"suite.sheets.trash.purge_trashed_sheets",
 		# mail
+		"suite.mail.retention.purge_expired_mail",
 		"suite.mail.doctype.jmap_account.jmap_account.delete_orphaned_jmap_accounts",
 		"suite.mail.doctype.mail_exchange.mail_exchange.clean_import_export_directories",
 		"suite.mail.doctype.push_subscription.push_subscription.renew_expiring_push_subscriptions",
--- a/suite/mail/retention.py
+++ b/suite/mail/retention.py
@@ -1,0 +1,209 @@
+# Copyright (c) 2026, Frappe Technologies Pvt. Ltd. and contributors
+# For license information, please see license.txt
+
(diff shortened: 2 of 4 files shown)


s-aga-r:
@krantheman Why explicit delete? Mail servers handle this implicitly.

https://stalw.art/docs/email/settings/maintenance/#auto-expunge

Screenshot 2026-07-30 at 8 32 50 AM

PR #410, opened 2026-07-30
fix(mail): keep the quoted reply trail out of message previews

krantheman:
Reported by Vibhav in #mail (Raven, 2026-07-29); Sagar confirmed previews are plain text.

**Bug:** The collapsed one-line preview of a mail in list rows includes the quoted reply trail, burying the actual new content.

**Cause:** `format_message` built the preview with `convert_html_to_text(html_body)` over the full HTML body — quotes included — then truncated to 256 chars.

**Fix:** Strip the `.gmail_quote` / `.frappe_mail_quote` containers (the same ones `EmailContent.vue` collapses client-side) before converting to text. If a message is nothing but a quote, fall back to the unstripped conversion so the preview never goes blank.

Before: `Sounds good, see you then! On 28 Jul 2026 at 4:47 PM, x@y.com wrote: the original ...`
After: `Sounds good, see you then!`

The `text_body` path is untouched — plain-text bodies carry no quote markers, so stripping there would need fragile heuristics.

### DIFF
--- a/suite/mail/doctype/mail_message/mail_message.py
+++ b/suite/mail/doctype/mail_message/mail_message.py
@@ -1124,6 +1124,19 @@ def fetch_blobs(account: str, blobs: list[str] | list[tuple[str, str | None]]) -
         frappe.throw(_("Failed to fetch blob(s)."))
 
 
+def preview_from_html(html_body: str) -> str:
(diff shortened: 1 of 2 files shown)


s-aga-r:
Preview is part of the email and cached in Data Store. So, this change will only apply to new emails; otherwise, we have to clear the cache for all.

PR #415, opened 2026-07-30
fix(mail): single-line rows in All Inboxes, which has no reading pane

krantheman:
Reported by Rushabh in #mail Raven (2026-07-30): All Inboxes renders tall three-line rows even though it has no reading pane.

**Cause:** `MailRow`'s `isFullWidth` keys off the global `show_reading_pane` setting, not off whether the enclosing view actually renders a pane — so with the setting on, the pane-less All Inboxes list stacked its rows as if a pane were eating two-thirds of the width.

**Fix:** a `no-reading-pane` prop threaded `AllInboxesView → MailListItem → MailRow`, folded into `isFullWidth`, so a pane-less view forces the single-line full-width layout on desktop. Mobile is untouched. All Inboxes is currently the only pane-less list — search and Screener render a pane when the setting is on — but any future one gets the same lever.

This is the interim half of Rushabh's ask; full split-view support in All Inboxes is tracked separately as an issue.

Before:
before

### DIFF
--- a/frontend/src/apps/mail/components/MailListItem.vue
+++ b/frontend/src/apps/mail/components/MailListItem.vue
@@ -4,6 +4,7 @@
 		:is-selected
 		:selectable
 		:selection-mode
+		:no-reading-pane
 		:unread="!mail.seen"
 		:hide-sender
 		:avatar-label="avatarLabel"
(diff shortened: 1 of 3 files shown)


krantheman:
replaced by https://github.com/frappe/suite/pull/424

PR #421, opened 2026-07-30
feat(mail): global screened senders with rebuild-all sieve action

s-aga-r:
## Problem

- Screening is enabled by default for new personal accounts, but the screened senders list has no knowledge of the server's local domains — so mail from colleagues on the same server lands in Screening.
- There was no way for an Administrator/System Manager to whitelist (or block) a domain or email address for **all** accounts.

## Solution

- A **Screened Email Address without an account is a global rule**: it applies to every account. Only a System Manager can manage global rules (enforced in `validate()`, since the unique index and permission query don't cover the account-less case; regular users never see global rows).
- The automation sieve is built from the **effective** list: global rules overlaid with the account's own, where the account's rule wins when both screen the same email or domain (`get_effective_screened_email_addresses`). `get_screened_email_addresses` stays account-only, so the settings UI never shows or edits the admin-managed globals.
- **Adding/removing a global rule rebuilds nothing** — an admin batches their changes and then triggers the new **Rebuild Automation Sieves** action in the Screened Email Address list view. It fans all JMAP accounts out into deduplicated long-queue batches with per-account error isolation, and rebuilds with `activate=False` so it never overrides a vacation auto-responder or a user-written active script.
- Duplicate global rules are caught in validation — the `(account, email)` unique index doesn't apply when account is NULL (MariaDB allows multiple NULL rows in a unique key).
- Fixed the stale `Blocked Email Address` doctype JS files left over from the rename.

With this, whitelisting local domains is: add global rules like `@yourcompany.com` → Accepted, then click **Rebuild Automation Sieves**.

## Tests

- `test_effective_screened_email_addresses_merge` — account rule overrides the global rule for the same value (case-insensitive).
- `test_duplicate_global_screened_email` — duplicate globals are rejected in validation.
- `test_global_rule_requires_system_manager` — non-admins cannot create global rules even with `ignore_permissions`.

All screened email address and sieve script tests pass. Existing sites need `bench migrate` to drop the mandatory constraint on `account`.

### DIFF
--- a/suite/mail/doctype/screened_email_address/screened_email_address.js
+++ b/suite/mail/doctype/screened_email_address/screened_email_address.js
@@ -1,4 +1,4 @@
 // Copyright (c) 2026, Frappe Technologies Pvt. Ltd. and contributors
 // For license information, please see license.txt
 
-frappe.ui.form.on('Blocked Email Address', {})
+frappe.ui.form.on('Screened Email Address', {})
--- a/suite/mail/doctype/screened_email_address/screened_email_address.json
+++ b/suite/mail/doctype/screened_email_address/screened_email_address.json
@@ -43,7 +43,7 @@
    "reqd": 1
   },
   {
-   "description": "The shared JMAP account this screened email belongs to. The screened list is the same for everyone with access to the account.",
+   "description": "The shared JMAP account this screened email belongs to. The screened list is the same for everyone with access to the account. Leave empty to make the rule global: it applies to every account, unless the account screens the same email or domain itself. Only a System Manager can manage global rules.",
    "fieldname": "account",
    "fieldtype": "Link",
    "in_list_view": 1,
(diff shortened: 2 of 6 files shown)


s-aga-r:
```py
from suite.mail.doctype.screened_email_address.screened_email_address import get_global_screened_email_addresses
from suite.mail.utils.validation import is_domain_entry

global_values = [row["email"] for row in get_global_screened_email_addresses()]

if global_values:
      doctype = frappe.qb.DocType("Screened Email Address")

      # Exact matches (emails and '@domain' rules), plus every address on a globally screened domain.
      email_matches = doctype.email.isin(global_values)
      for value in global_values:
              if is_domain_entry(value):
                      email_matches |= doctype.email.like(f"%{value}")

      (
              frappe.qb.from_(doctype)
              .delete()
              .where(
                      email_matches
                      & doctype.account.isnotnull()
                      & (doctype.account != "")
                      & (doctype.action == "Accepted")
              )
              .run()
      )
```

PR #441, opened 2026-07-31
fix(drive): post-migration cleanup, plus two Writer UX fixes

safwansamsudeen:
Follow-ups after the flat single-root Drive migration, plus two Writer fixes that landed on the same branch.

## Drive
- Fold Shared into Home and rename Site to Everyone; drop the site-access marker on the shared root
- Let admins create in the shared root; don't leave an empty Previous Teams behind
- Drop site-wide storage from Statistics
- Open attachment nodes instead of routing them as folders; point the Attachments crumb at the section, not the page
- Let the nightly purge actually reclaim removed files

## Writer
- The tab context menu (Rename / Copy Link / Delete) is now also reachable from a vertical triple-dot button on the active tab, not only by right-click
- Restored the tooltip on the Custom Spacing toolbar button — `MenuItems.vue` only wraps plain command items in a `Tooltip`, so a component item has to render its own trigger

## Testing
`specs/writer` + `specs/drive-writer` pass locally (17/17). New coverage: renaming a tab through the triple-dot menu, and the Custom Spacing tooltip asserted together with the popover still opening (wrapping the trigger slot in a `Tooltip` breaks `PopoverTrigger`'s as-child wiring).


safwansamsudeen:
Merging with CI failure as I need to deploy fixes now 🥲.

PR #477, opened 2026-08-04
feat(sheets): peer-to-peer collab transport, no server to deploy

AsifMulani1:
Adds a third collab transport for Sheets, selected by `collab_mode` in
site_config. **Off by default** — no site changes behaviour until that key is
set, so this is safe to merge ahead of deciding to enable it.

| mode | transport |
|---|---|
| `realtime` *(default)* | existing `frappe.realtime` relay, unchanged |
| `webrtc` | peer-to-peer via `signal.frappe.cloud` — the transport Suite's Writer already uses |
| `hocuspocus` | existing websocket path, needs a collab server |

## Why

Sheets' Hocuspocus path needs a stateful Node process nobody is running.
Writer solves the same problem with no per-product infrastructure at all, and
Frappe already operates the signaling + TURN pair it points at. This takes
that route.

## Two deliberate departures from Writer

**Rooms are encrypted.** Writer's rooms are open — the signaling server
authenticates nobody, so anyone who knows a document id can join and read the
live document. `get_collab_session` derives a per-sheet key from the site
secret and hands it only to callers who pass a read check, so an uninvited
peer sees ciphertext.

**Saves merge instead of overwriting.** Writer stores whatever the last saver
sent, which is only safe while every peer is connected P2P and therefore holds
a superset of the edits. `save_collab_state` applies the incoming update onto
the stored one with pycrdt, so a peer stranded behind an untraversable NAT
can't clobber edits it never received.

## Room keys are revocable

The key is derived, so without care it would be a permanent grant. A
`room_epoch` counter on `Sheet Collab State` feeds the derivation; unsharing
bumps it, and so does a write→read downgrade. Granting deliberately does not —
rotating splits a live session until every peer reloads, which is the point of
a revocation and pure disruption for a grant.

## Known limitation

P2P cannot enforce read-only downstream: any peer can transmit to any other, so
a viewer who joins can publish into the live session. `save_collab_state`
re-checks write permission, so nothing they send persists. This matches Suite's
Writer, which guards its durable writes (`check_permission("write")` on
`save_doc` and friends) while leaving the live channel open. Documented in
`webrtc-client.js`. If we later want read-only enforced live, that requires a
mediator — i.e. the Hocuspocus path.

## ⚠️ Requires `bench migrate`

`room_epoch` is a new field on `Sheet Collab State`. Without the migration
`get_collab_session` hits an unknown column and fails.

## Verification

Unit: 95 backend, 1115 frontend, ruff clean.

End-to-end against a running bench, each peer in an isolated browser context —
separate cookies *and* separate IndexedDB, so nothing can pass from shared
y-indexeddb cache:

- three peers editing different cells at once → all three converge, no loss
- three peers writing the **same** cell at once → all agree on one value (no split brain)
- a late joiner reconstructs the full prior state
- every peer's edit reaches the backend merged into one Y.Doc
- no collab console errors on any peer

Revocation exercised against the real database (in a rolled-back transaction):
granting write leaves the key alone; downgrade and unshare both rotate it;
unsharing a user who had no share doesn't churn.

## Rollback

Remove `collab_mode` from site_config. No migration to reverse, no
infrastructure to tear down.

### DIFF
--- a/frontend/src/apps/sheets/collab/persistence.js
+++ b/frontend/src/apps/sheets/collab/persistence.js
@@ -1,0 +1,147 @@
+// Y.Doc ⇄ Frappe persistence for the peer-to-peer collab path.
+//
+// In the websocket design the collab server owned persistence: it held the
+// authoritative doc and flushed it on a debounce. With P2P there is no server
+// in the data plane, so the browsers are responsible for getting state back to
+// Frappe — the same shape Writer uses, where `useYjs`'s `save()` posts the
+// encoded doc from whichever client happens to be editing.
+//
+// Every writer flushes its own view of the document. That's safe because the
+// backend merges instead of overwriting (see `sheets.collab.save_collab_state`),
+// so two peers flushing concurrently union rather than clobber. It also means a
+// lost save costs nothing: the next debounce window sends the full state again.
+//
+// Viewers never flush at all — `canWrite` gates the whole path, and the
+// endpoint re-checks write permission regardless.
+
+import * as Y from 'yjs'
+
+// Writer debounces its save at 5s. Matching it keeps the request rate
+// predictable for a shared backend: one POST per active editor per 5s of
+// continuous typing, plus a flush when the editor closes.
+export const SAVE_DEBOUNCE_MS = 5000
+
+// `btoa` needs a binary string, and spreading a large Uint8Array into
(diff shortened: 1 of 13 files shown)


BreadGenie:
Closing this since we will implement a common solution for all drive backed apps.

PR #480, opened 2026-08-04
feat(mail): reliable dark-mode email rendering

### LINKED ISSUE #405: [Bug]: Share notification emails are unreadable in dark mode and unstyled in light
`suite/templates/emails/drive_share.html` is a standalone fragment: it doesn't extend `templates/emails/base.html`, hardcodes `#444`/`#888` text on a transparent background and a `#171717` button. In dark-mode clients the text is dark-on-dark and the CTA invisible; in light clients the CTA renders as a blank dark pill. Mail's share/notification path uses the same macros (per discussion in Raven).

dark mode

light mode

Related but distinct from #42/#57 — those are about rendering third-party mail; this is our own outbound templates. Standardized email templates were requested from design in the same thread.

Reported by Rushabh in the #mail Raven channel (2026-07-28).

krantheman:
Replaces the dark-mode text-walker heuristic with a deterministic color remap and fixes the outbound drive templates (#405).

**Reader (`darkMail.ts` + `EmailContent.vue`)**
- Every color literal (inline styles, `<style>` sheets, `bgcolor`/`color` attrs) goes through one OKLab mapping — no computed styles, no ancestor heuristics. Backgrounds invert end-to-end so surfaces that contrasted in light still contrast in dark; gray text flips to a light band and sheds slate tints; colorful text keeps its hue, brightened only to a WCAG AA floor.
- Text on a raster background (`url(...)`) keeps its authored color — the image's pixels don't change in dark mode, so remapped text would contrast against the wrong surface (dark heading flipped near-white over a light gradient JPEG). Nearest surface decides: color-backed cards nested in an image hero still remap.
- Author `@media (prefers-color-scheme: dark)` blocks are dropped and pure-light blocks unwrapped before judging or remapping: the srcdoc iframe inherits the app's scheme so those blocks *do* activate, but sanitization strips the attributes their selectors key on (Discourse's `dm="body"`), leaving half a dark design — a broad `color: inherit !important` with none of its surfaces. Every email is remapped as its light-scheme self, the one design that's always fully authored.
- Art-directed emails (full-bleed, content-holding, colored canvas) render as authored. Hidden preheader text (authored in its surface's color) stays hidden. Quoted content can't claim a reply's canvas.
- Sanitizer now allowlists `colspan`/`rowspan` — stripping them collapsed table-email grids.
- 36 unit tests on the mapper and detectors.

**Outbound (fixes #405)**
- `drive_share` / `drive_invitation` rebuilt on the calendar card pattern: explicit colors on every surface, white CTA label (no more blank dark pill in light clients), Frappe Drive wordmark with CID-embedded logo.

Also: thread header keeps its gap above image-only mails; upcoming-events count shows only when events are hidden below the fold.


krantheman:
@greptile review

PR #481, opened 2026-08-04
fix(drive): mark notifications read when files open

BreadGenie:
shared drive files when open not from notification list doesn't get marked as read


BreadGenie:
broken test fixed in #483 merging it

PR #483, opened 2026-08-04
fix(drive): restore permission aware attachment listings

BreadGenie:


BreadGenie:
ci failure was due to cache merging this for now

PR #485, opened 2026-08-04
fix(deps): widen beautifulsoup4 pin to resolve with frappe

Delte:
## Problem

`suite`'s `pyproject.toml` pins `beautifulsoup4~=4.15.0`, while `frappe`
pins `beautifulsoup4~=4.13.5`. These two ranges never overlap
(`~=4.13.5` means `>=4.13.5,<4.14.dev0`; `~=4.15.0` means
`>=4.15.0,<4.16.dev0`), so any strict dependency resolver that solves
`frappe` and `suite` together (e.g. `uv`, as used by `pilot get-app`)
fails with an unsatisfiable-dependencies error:

```
Error: 'suite' has dependencies that can't be resolved
  beautifulsoup4>=4.15.0,<4.16.dev0 and beautifulsoup4>=4.13.5,<4.14.dev0
```

Classic `bench`/`pip` doesn't do a unified strict resolve across all
installed apps, so this conflict doesn't surface there — but it's a
real, pre-existing inconsistency between the two repos' declared
dependencies, not something introduced by installing newer versions
of either app. It reproduces at the exact commit pair we validated
during a migration (`frappe@06613fc6`, `suite@1c8b28e2`), and still
reproduces at both repos' current `develop` heads.

## Fix

Widen `suite`'s pin to a range that includes `frappe`'s pin:

```diff
- "beautifulsoup4~=4.15.0",
+ "beautifulsoup4>=4.13.5,<5",
```

No code changes were needed elsewhere — `suite`'s own bs4 usage (in
`mail`/`client`, per the comment on that dependency block) doesn't
rely on anything introduced between 4.13.5 and 4.15.0 as far as I
could tell from a quick read; happy to narrow the upper bound if
there's a known reason for the `4.15.0` floor that I'm missing.

Tested by installing `frappe` + `suite` together via `pilot get-app`
(which runs `uv`'s strict resolver) — resolves cleanly with this
change, fails without it.


s-aga-r:
Thanks @Delte for the fix! Your dependency commit has been carried over (with your authorship preserved) to #512 — closing this one since the branch also contains an unrelated commit (847aa6cd1) that we don't want to merge.

Delte:
i will open another PR for variable font names, it wasn't intend to be in same PR

PR #505, opened 2026-08-06
fix(mail): raise on JMAP method-level errors in changes calls

s-aga-r:
## Problem

`fetch_changes` crashed with a misleading `KeyError: 'created'` when the JMAP server answered `Email/changes` with a method-level error response:

```
result = {'type': 'forbidden', 'description': 'You are not authorized to perform this action'}
builtins.KeyError: 'created'
```

`EmailService.changes()` (and the equivalent `changes()` wrapper in the other seven services) unwraps the first method response as if it were always a changes result, so an `["error", {...}]` response flowed downstream as-is.

## Fix

- `CoreService._changes()` now raises a `RuntimeError` naming the JMAP error (e.g. `Email/changes failed: {'type': 'forbidden', ...}`) when the server returns an `error` method response. The check lives in the shared base method so all services with a `changes()` wrapper are covered.
- `fetch_changes` bails out with a warning if `changes()` returns an empty dict (the no-`methodResponses` path), which would previously hit the same `KeyError`.

The failure is still logged via the existing handler in `fetch_changes`, but as a readable error naming the actual server response, and the sync state is left untouched.

### DIFF
--- a/suite/mail/doctype/mail_message/mail_message.py
+++ b/suite/mail/doctype/mail_message/mail_message.py
@@ -1365,6 +1365,10 @@ def fetch_changes(user: str, account: str, email_state: str | None = None, ctx: 
 
         result = email_service.changes(current_state)
 
+        if not result:
+            logger.warning("empty-changes-response")
+            return
+
         if created_ids := result["created"]:
             logger.info("new-messages-created", count=len(created_ids))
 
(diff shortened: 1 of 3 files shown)


s-aga-r:
@greptileai re-review

PR #517, opened 2026-08-06
feat(writer): added DOCX import support

### LINKED ISSUE #516: [Feature]: Import DOCX files into Writer
### App

Writer

### Problem / motivation

Writer can export a document as DOCX, but there's no way to bring one back in. If you have an existing Word document, the only option is to copy-paste it, which loses images and mangles tables. This makes it hard to move existing work into Writer, or to round-trip a document that was exported earlier.

### Proposed solution

Add an "Import → DOCX" option to the document menu, next to the existing Export option.
Picking a .docx would convert it and load the content into Writer, keeping headings, lists, tables and images. Images would be uploaded as normal attachments rather than embedded inline. If the current document is empty, the content loads into it. Otherwise it opens in a new tab, so existing work isn't touched.
Conversion can happen fully in the browser (e.g. mammoth), the same way Sheets already does its import with xlsx — so no new backend endpoint is needed.

DikshaJadhav3110:
## Summary

This PR adds **Import → DOCX** support to Writer, allowing users to import Word (`.docx`) documents directly into the editor.

If the current document is empty, the imported content replaces it. Otherwise, the content is imported into a new tab to avoid overwriting existing work.

Closes #516 

## Implementation

- Added **Import → DOCX** to the Writer document menu.
- Uses **mammoth.js** for client-side DOCX parsing, so no backend conversion endpoint is required.
- Uploaded images through the existing `suite.writer.api.embed.add` API, reusing the same upload flow as the editor.
- Normalized imported tables to match Writer's native table structure, ensuring consistent editing and styling.
- Preserved common formatting including headings, lists, tables, links, images, paragraphs, and inline formatting.
- Fixed the existing **Create Tab** behavior so creating a tab now preserves existing content and opens a new empty tab.

## Verification

Verified manually by importing multiple `.docx` files containing:

- Headings
- Paragraphs
- Ordered and unordered lists
- Tables
- Hyperlinks
- Images
- Mixed formatting

Also verified that:

- Importing into an empty document replaces its contents.
- Importing into a non-empty document creates a new tab.
- Imported images are uploaded correctly.
- Imported tables remain editable and match Writer's native table behavior.
- No console errors were observed during import.

## Notes

- Table formatting is normalized to Writer's table model, so imported tables behave the same as tables created inside Writer.
- Headers and footers (including images in them) are not imported; mammoth does not read those parts of a .docx and Writer has no header/footer to map them onto.

https://github.com/user-attachments/assets/594056ea-b8cb-4637-bcdd-a0b449d8dca4


BreadGenie:
Try to add demos instead of a long PR description. It would make it much easier to review the PR. 
> A ~picture~ gif is worth a thousand words.

DikshaJadhav3110:
Thanks for the feedback @BreadGenie. I've added a demo video to the PR so the changes are easier to review, and I'll keep future PR descriptions more concise as well.

PR #519, opened 2026-08-06
fix(writer): marking a document as favourite in Writer sync to Drive's Favourites tab

### LINKED ISSUE #518: [Bug]: Marking a document as favourite in Writer doesn't sync to Drive's Favourites tab
### App

Writer

### What happened?

Marking a document as favourite from the Writer navbar (⋯ menu → Favourite) doesn't sync with Drive. The star appears to toggle in Writer, but:

- The document never appears in Drive's Favourites tab.
- Reloading the page reverts the favourite state — it isn't persisted.

https://github.com/user-attachments/assets/487bb279-0caf-417a-9993-a984e12449ea

### Steps to reproduce

Open any document in Writer.
Click the "⋯" menu in the navbar → select "Favourite".
Reload the page — the star reverts to unfavourited.
Go to Drive → Favourites tab — the document isn't listed.

### Environment

Any browser

DikshaJadhav3110:
### Summary

Favouriting a document from Writer's navbar (⋯ → Favourite) never synced to Drive:

- The document didn't appear in Drive's Favourites tab.
- The favourite state didn't survive a page reload.
- After a partial fix, favouriting would also break the editor (page became read-only until reload).

### Root cause
Writer called `toggle_favourite`, a `File `doc-method, through Frappe's generic doc-method dispatcher. That dispatcher auto-attaches the raw document (`doc.as_dict()`) to every response. Frappe-ui's global fetch layer uses that to fully overwrite the client's cached document — wiping out fields like `write/share` that are only populated by the enriched fetch, which is what broke editing.

The underlying `toggle_favourite` method also lacked `@frappe.whitelist()` (dropped accidentally in an earlier refactor), and separately was being called with an argument shape (`entities: [...]`) it didn't accept.

### Fix
Instead of patching that path further, Writer now calls the same shared, whitelisted API Drive's own UI already uses — `suite.drive.api.files.set_favourite`, via the existing `toggleFav` resource in `@/apps/drive/resources/files`. This is a plain `/api/method` call, so it never touches the doc-method dispatcher and never clobbers the cached document.

- `suite/drive/overrides/file.py`: removed the now-unused t`oggle_favourite` doc-method.
- `frontend/src/apps/writer/composables/useDocument.ts`: removed the doc-method wiring that pointed at it.
- `frontend/src/apps/writer/components/Navbar.vue`: Favourite/Unfavourite now call the shared `toggleFav` resource, same as Drive.

Closes #518 

https://github.com/user-attachments/assets/22f004fd-2d48-424a-92b6-827e777274c3

### DIFF
--- a/frontend/src/apps/drive/resources/files.js
+++ b/frontend/src/apps/drive/resources/files.js
@@ -197,6 +197,22 @@ export const toggleFav = createResource({
       toast(`${toggleFav.params.entities.length} items unfavourited`)
     else toast(`${toggleFav.params.entities.length} items favourited`)
   },
+  onError() {
+    if (toggleFav.params?.entities) {
+      const reverted = toggleFav.params.entities.map((e) => ({
+        ...e,
+        is_favourite: !e.is_favourite,
+      }))
+      getFavourites.setData((d) => {
+        const names = reverted.map((r) => r.name)
+        return reverted[0].is_favourite
+          ? [...(d ?? []), ...reverted]
+          : (d ?? []).filter(({ name }) => !names.includes(name))
+      })
+      mutate(reverted, (el, { is_favourite }) => (el.is_favourite = is_favourite))
+    }
+    toast.error('Failed to update favourite status')
+  },
 })
 
 export const clearRecent = createResource({
--- a/frontend/src/apps/writer/components/Navbar.vue
+++ b/frontend/src/apps/writer/components/Navbar.vue
@@ -239,11 +239,18 @@ const fileActions = computed(() =>
               icon: LucideStar,
               onClick: () => {
                 props.file.doc.is_favourite = true
-                toggleFav.submit({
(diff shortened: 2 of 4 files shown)


BreadGenie:
@DikshaJadhav3110 can you address the concerns by greptile?

BreadGenie:
why was the toggle favorite method removed? what replaces it?

DikshaJadhav3110:
toggle_favourite was removed because it was being called as a Frappe doc-method. That API automatically returned the full File document, which replaced Writer's cached document and removed fields like write, share, and breadcrumbs. This caused parts of the Writer UI to break until reload.
Instead, Writer now uses the existing: `suite.drive.api.files.set_favourite`.
This is a normal whitelisted API, so it performs the favourite operation without replacing the cached document.
It also means Writer and Drive now use the same favourite API, avoiding duplicate logic and keeping both implementations in sync.

PR #520, opened 2026-08-07
fix(writer): prevent metadata leakage in search response

DikshaJadhav3110:
The Writer search API correctly removes documents that the user doesn't have permission to access. However, it still returns the search summary (match counts and spelling suggestions) without applying the same permission check.

Because of this, a user searching for something in a private document could still learn that the document exists, and in some cases even discover confidential words through the spelling suggestions, even though the search results are empty.

This fix updates the summary after permission filtering, so the match counts only reflect documents the user can actually access, and removes the unused spelling-correction fields to avoid leaking information.


BreadGenie (changes requested):
can you setup pre-commit locally so that linters won't trip in CI?

DikshaJadhav3110:
Pre-commit is now set up locally.
The changes from this PR overlap with #622, so I've combined them there to keep everything in one PR.

Closing this PR in favor of #622.

PR #529, opened 2026-08-07
fix: move blob store to site directory root

s-aga-r:
## Problem

The blob store was recently moved from `sites/<site>/private/files/blob-store` to `sites/<site>/private/blob-store` to keep it out of files backups. However, Frappe Cloud backs up the **entire** `private` directory, so the blob cache still ends up in every site backup.

## Fix

- Move the blob store to the site directory root: `sites/<site>/blob-store`. It remains per-site (multi-tenant safe) and is never web-served, but is no longer included in backups.
- Add a `move_blob_store_out_of_private` patch that relocates `private/blob-store` to the site root with the same best-effort semantics as the earlier relocation patch (if the destination already exists, the old directory is dropped rather than merged — blobs are a cache refetched on demand).
- Sites still on the oldest layout (`private/files/blob-store`) are moved straight to the site root by the existing `move_blob_store_out_of_files` patch, since it resolves the destination via `get_blob_base_path()`; the new patch then finds nothing to do.

## Verification

Ran the patch on a local site: `private/blob-store` (with the `mail` namespace) was moved to `sites/<site>/blob-store` intact, and `get_blob_base_path()` resolves to the new location.

### DIFF
--- a/suite/patches.txt
+++ b/suite/patches.txt
@@ -95,3 +95,4 @@ suite.mail.patches.reset_push_subscription_types_to_all
 # --- suite_core ---
 # Existing, already-set-up sites shouldn't be sent through Suite's onboarding.
 execute:frappe.db.set_single_value("Suite Settings", "is_onboarded", 1) if frappe.is_setup_complete() else None
+suite.suite_core.patches.move_blob_store_out_of_private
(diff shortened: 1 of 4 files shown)


s-aga-r:
@greptileai review

PR #545, opened 2026-08-08
fix(calendar): invite mailing list members individually

s-aga-r:
## Problem

An invitation sent to a mailing list (`team@example.com`) reaches every member's inbox, but the event is never added to anyone's calendar. The same invitation sent to an individual or a group works.

This is by design in Stalwart rather than a bug, and it falls out of how the two objects are modelled:

- A mailing list resolves to `RcptResolution::Expand(list.recipients)` at the SMTP RCPT stage. The list address is popped off `rcpt_to` and survives only as the DSN `ORCPT` string, so each member's copy is delivered to their own address.
- The iTIP ingest then matches the iCalendar `ATTENDEE` against the recipient account's own addresses, which are built from the account's aliases plus the addresses of every group in `member_group_ids`. A mailing list stores its members as plain address strings with no back-reference to the member accounts, so it can never appear there.

The attendee therefore matches nobody, ingest returns `NotOrganizerNorAttendee`, and nothing is written to any calendar. Groups are unaffected because a group is a real principal with its own address and its own calendar.

Switching the affected addresses to groups is not an option: a list delivers a copy to each member's inbox, whereas a group is a shared mailbox, and that difference is usually the reason the list exists.

## Fix

Expand a mailing list into one participant per member before the event is stored, which is what RFC 5546 asks the organizer to do.

Doing it at the point the participants are persisted (rather than when the invitation mail is built) matters, because Suite has two invite paths and both are built from the stored participant list: its own templated mails when `custom_event_invites` is on, and the JMAP server's scheduling messages otherwise. Each member also ends up with a distinct participant uid, which is what gives them individual RSVP links.

Details:

- Addresses resolve via a new `MailingListService.get_address_index()`, covering each list's primary address and its **enabled** aliases — a disabled alias is not delivered to, so an invite to it must not expand either. Cached for 60s alongside `get_domains`.
- Nested lists are flattened in place, with a visited set that also terminates membership cycles.
- A member invited separately is not duplicated; the explicit entry wins, so a response already recorded against it is preserved.
- The directory lookup is skipped entirely when no participant sits on a local domain.
- If the directory is unreachable, participants pass through unchanged — an event is not worth failing over.
- Participants are re-expanded on every update, since membership is only a snapshot taken when the invitation is sent.

Controlled by `expand_mailing_list_participants` and `max_mailing_list_participants`, both read through `get_config` so either site config or Mail Settings can supply them. A patch enables them on existing sites, matching the `custom_event_invites` precedent.

## Known limitations

- **External organizers are unchanged.** Someone on Gmail or Outlook inviting a list hits the same wall, because the expansion happens inside Stalwart at RCPT time and the list address is gone before ingest runs. Fixing that needs a server-side change.
- **Membership is a snapshot.** Someone added to the list later receives list mail but not this event until the organizer touches it again.
- **Expansion discloses membership.** It runs with the app's Stalwart credentials rather than the acting user's, so anyone who can create an event can see a list's members by inviting it. A per-list opt-in would be the natural follow-up if that matters for lists like `all@`.
- Once expanded, the event shows individual members rather than the list address, and every attendee can see the other attendees.

## Tests

- 16 in `test_calendar_mailing_list_participants.py` — expansion against a fixed directory (ordering, dedupe, nested lists, cycles, inherited roles, the size cap, the off switch, unreachable directory), the `get_config` wiring, plus an end-to-end case that creates a real mailing list with two members, invites the list address, and asserts the stored event carries both members with distinct uids and no trace of the list address.
- One added to `test_admin_groups_and_lists.py` proving the index resolves a real list's primary address and enabled alias while excluding a disabled one.

`test_calendar_events`, `test_calendar_invites_rsvp`, `test_calendar_mail_invites`, `test_calendar_calendars`, `test_admin_groups_and_lists`, `test_mail_flags_and_search` and `test_mail_misc` all pass.

## Second commit

`chore(mail): use dummy domains and names in tests and examples` is unrelated cleanup, kept separate for review: it replaces `frappe.io` and real people's names in the mail test fixtures (and the comments those tests document) with RFC 2606 reserved domains and placeholder names.

### DIFF
--- a/suite/calendar/doctype/calendar_event/calendar_event.py
+++ b/suite/calendar/doctype/calendar_event/calendar_event.py
@@ -20,6 +20,7 @@ from suite.calendar.doctype.calendar_event.invitations import (
     acting_as_organizer,
     custom_event_invites_enabled,
 )
+from suite.calendar.doctype.calendar_event.mailing_lists import expand_mailing_list_participants
 from suite.mail.doctype.user_account.user_account import get_user_for_jmap_account
 from suite.mail.jmap import get_calendar_event_service, get_jmap_connection
 from suite.mail.jmap.services.calendars.calendar_event import CalendarEventService
@@ -369,6 +370,7 @@ def add_calendar_event(
 
     uid = uuid7().hex
     creation_id = str(uuid7())
+    participants = expand_mailing_list_participants(participants)
     event = {
         "creation_id": creation_id,
         "uid": uid,
@@ -481,6 +483,7 @@ def update_calendar_event(
 ) -> None:
     """Updates a calendar event for the given account and event ID."""
 
+    participants = expand_mailing_list_participants(participants)
     event = {
         "id": id,
         "uid": uid,
@@ -539,6 +542,9 @@ def update_calendar_event_instance(
     send_scheduling_messages: bool = False,
 ) -> None:
     """Updates a specific instance of a recurring calendar event based on its master ID and recurrence ID."""
+
+    if "participants" in patch:
+        patch = patch | {"participants": expand_mailing_list_participants(patch["participants"])}
 
     use_custom_invites = False
     next_sequence = None
--- a/suite/calendar/doctype/calendar_event/mailing_lists.py
+++ b/suite/calendar/doctype/calendar_event/mailing_lists.py
@@ -34,6 +34,14 @@ def expand_mailing_list_participants(participants: list[dict] | None) -> list[di
     Participants that are not mailing lists are passed through untouched, and the original order is
     preserved. Returns the input unchanged when expansion is disabled or the directory cannot be
     reached — a calendar event is never worth failing over this.
+
+    The size cap bounds the event's total participants, but only members added by expansion are ever
+    dropped to honour it. A participant the organizer named themselves is always kept, even when
+    that pushes the total past the cap, because dropping one removes them from the stored
+    event, and on the next update the invitation code reads that as a withdrawn attendee and mails
+    them a cancellation. For the same reason an explicit participant always wins over the same
+    address arriving through a list, no matter which order they appear in: the explicit entry
+    carries the uid their RSVP is recorded against, and a member entry would reset it.
     """
 
     if not participants or not _expansion_enabled():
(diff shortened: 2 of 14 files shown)


s-aga-r:
@greptileai review

s-aga-r:
### Limitation: this only covers invites created here, not ones arriving from outside

Worth stating plainly, since the behaviour differs depending on where the invite originates.

**Works.** An event created in Suite with a mailing list address among its participants. The list is expanded into its members before the event is stored, so every member ends up with the event on their calendar and their own RSVP link.

**Still does not work.** An invite sent to a mailing list address from outside this server — Gmail, Outlook, another Suite deployment, any external MTA. Members receive the invitation mail in their inbox exactly as before, but the event is **not** added to any of their calendars.

**Why the two differ.** This PR expands the list at the point the organizer saves the event, which is only reachable for invites that originate here. For inbound mail there is no such point: Stalwart resolves the list at the SMTP RCPT stage, where `team@example.com` is replaced by each member's own address and the list address is popped off `rcpt_to`, surviving only as the DSN `ORCPT` string. By the time the message reaches iTIP ingest, the recipient is the individual member while the iCalendar `ATTENDEE` still names the list. Ingest matches that attendee against the account's own addresses — its aliases plus the addresses of the groups in `member_group_ids` — and a mailing list can never appear there, because it stores its members as plain address strings with no back-reference to the accounts. No attendee matches, so nothing is written to any calendar.

**How to confirm you are hitting this.** The failure is logged rather than silent. Each member's delivery emits a `Calendar(ItipMessageError)` event with details `Not an organizer or attendee`.

**What a fix would need.** A server-side change, not an app-side one. The queue's `Recipient` already carries the pre-expansion list address in its `orcpt` field, but only `rcpt.address` is passed to the ingest, so threading `orcpt` through and treating it as an address of the receiving account would make the attendee match. That alone is not sufficient though: every member would then hold an event whose `ATTENDEE` is the list, so their replies would all come back under the same identity. Doing it properly means rewriting the attendee to the receiving member on import, which is a real piece of work rather than a small patch.

**In the meantime.** For invitations that must reach calendars, the organizer needs to address the members directly, or use a group — accepting that a group is a shared mailbox rather than a copy in each member's inbox, which is usually the reason the list exists in the first place.

PR #553, opened 2026-08-09
feat(calendar): add rate limits

s-aga-r:
Calendar has no rate limits today — there is no `calendar/install.py` at all. Creating or editing an event mails an invitation to every participant, and a mailing list participant expands to many, so it's the same fan-out that `outbound.send` is limited for. This adds `suite/calendar/install.py` on the same shape as the mail one and wires it into the `suite_core.boot.after_install` dispatcher.

Sibling of #551 (mail).

### Sizing

Unlike the mail APIs these endpoints are all interactive, and every person in an office can share one public address, so the windows are **per minute**: a runaway script trips them within seconds, while a floor of people working a calendar never gets close. Event creation carries an hourly ceiling on top, since that is the endpoint that fans invitation mail out to arbitrary addresses.

| Endpoint | Limit | Why |
|---|---|---|
| `calendar_event.add_calendar_event` | 60/min + 600/h | Mails an invitation to every participant; the hourly ceiling is the anti-spam backstop. |
| `calendar_event.update_calendar_event` | 120/min | Most frequent write (dragging an event reschedules it), each can send an update to every participant. |
| `calendar_event.update_calendar_event_instance` | 120/min | Same, single occurrence. |
| `calendar_event.delete_calendar_events` | 60/min | Sends a cancellation to every participant. |
| `calendar_event.delete_calendar_event_instance` | 60/min | Same, single occurrence. |
| `api.edit_calendar_event` | 120/min | The path the frontend uses; delegates to `update_calendar_event`, so an edit spends from that limit too — which makes the doctype-level one the aggregate across both entry points. This only holds with the limiter fix below. |
| `api.rsvp_calendar_event` | 60/min | Writes the response and emails the organizer. |
| `api.invites.rsvp_to_invite` | 60/min | Puts the event on the calendar, then the above. |
| `api.invites.add_invite_to_calendar` | 60/min | Parses a `text/calendar` blob and creates events. |
| `api.invites.get_invite_details` | 120/min | Parses a blob plus a JMAP lookup on every thread open. |
| `calendar.add_calendar` | 20/min | Calendars are created once and then lived in. |
| `calendar.update_calendar` | 60/min | |
| `calendar.delete_calendars` | 20/min | |

Endpoints that need the `@dynamic_rate_limit` decorator to accept a limit got it.

### Also fixes: limits were keyed by the request, not by the method

The nesting above surfaced a bug in `dynamic_rate_limit`, so a second commit fixes it. The wrapper resolved its bucket from `frappe.form_dict.cmd`, which names the request rather than the function it decorates:

- **Nested endpoints charged the wrong bucket.** `edit_calendar_event` → `update_calendar_event` charged the *edit* bucket twice and never touched the update one — halving edit's effective limit and leaving the aggregate uncounted. Every other nesting was hit too: `bulk_delete` → `delete_calendar_events`, and Meet's scheduler → `add_calendar_event`, both resolved to a caller with no rows at all, so the limit was skipped outright.
- **Legacy aliases were unlimited.** `override_whitelisted_methods` maps `mail.api.outbound.send` → the real path but leaves the alias in `cmd`, so no row matched and every limit was skipped. No Rate Limit row can name the alias either — `validate_method_path` resolves through `frappe.get_attr`, and an alias is not importable.
- **v2 requests shared one bucket.** With `cmd` unset the fallback path was already correct for lookup, but the counter key still landed wherever the outer call had published.

Measured on a site, one dispatched call through the outer endpoint, counters read straight out of redis:

| dispatch | before (outer / inner) | after |
|---|---|---|
| v1, real path in `cmd` | 2 / 0 | 1 / 1 |
| v2, `cmd` unset | 2 / 0 | 1 / 1 |
| legacy alias in `cmd` | 0 / 0 | 1 / 1 |

The fix resolves the bucket from the declared path and publishes that path while the call runs. This closes follow-up 2 from #551.

### Deliberately skipped

- **The guest RSVP page** (`/event_rsvp`). It is the only unauthenticated calendar surface and it does write through JMAP, but every recipient of one invitation can sit behind a single corporate NAT — a per-IP limit there would reject genuine responses. Replay of a valid signed token is the only real abuse case, and it is idempotent.
- **Read endpoints** (`get_calendars`, `get_calendar_events`, `fetch_calendar_events`, `fetch_calendars`, event notification reads). They fire on every calendar navigation, and a shared office address would trip them long before an abuser would — same call as `search_mails` / `get_avatar` in #551.
- **`CalendarExchange.retry`.** It re-enqueues a heavy job and is worth limiting, but it is a document method: `frappe.get_attr` can't resolve `...calendar_exchange.CalendarExchange.retry`, so the Rate Limit doctype can't reference it. Would need the mechanism to grow support for doc methods (Mail Exchange has the same gap).

### Verified

- All 13 method paths resolve and carry `@dynamic_rate_limit` (a missing decorator makes `create_rate_limit` throw at install time).
- Ran `add_rate_limits()` against a site inside a transaction: all 14 rows insert, and both windows on `add_calendar_event` come back from `get_rate_limits`.
- All 39 calendar tests pass, plus `suite.meet.api.test.test_meeting` — Meet schedules through `add_calendar_event`, so it shares that limit.
- The three dispatch shapes above are now covered by tests in `test_rate_limit.py` (the doctype had none). Each one fails on the unfixed limiter with exactly the counts in the table.

### DIFF
--- a/suite/calendar/api/__init__.py
+++ b/suite/calendar/api/__init__.py
@@ -14,6 +14,7 @@ from suite.calendar.doctype.calendar_event.calendar_event import (
 )
 from suite.mail.jmap import get_calendar_event_service
 from suite.mail.utils.dt import normalize_utc_z
+from suite.utils.rate_limiter import dynamic_rate_limit
 
 
 @frappe.whitelist()
@@ -124,6 +125,7 @@ def _with_name(items: list[dict] | None) -> list[dict] | None:
 
 
 @frappe.whitelist()
+@dynamic_rate_limit()
 def rsvp_calendar_event(account: str, id: str, response: str) -> None:
     """Records the logged-in user's RSVP (accepted / declined / tentative) on the event.
 
@@ -135,6 +137,7 @@ def rsvp_calendar_event(account: str, id: str, response: str) -> None:
 
 
 @frappe.whitelist()
+@dynamic_rate_limit()
 def edit_calendar_event(account: str, id: str, **kwargs) -> None:
     events = get_calendar_events_by_ids(account, [id])
     if not events:
--- a/suite/calendar/api/invites.py
+++ b/suite/calendar/api/invites.py
@@ -12,9 +12,11 @@ from suite.calendar.doctype.calendar_exchange.calendar_exchange import SERVER_MA
 from suite.mail.jmap import format_jmap_error, get_calendar_event_service, get_participant_identities
 from suite.mail.jmap.services.calendars.calendar import CalendarService
 from suite.mail.jmap.services.calendars.calendar_event import CalendarEventService
+from suite.utils.rate_limiter import dynamic_rate_limit
 
 
 @frappe.whitelist()
+@dynamic_rate_limit()
 def get_invite_details(account: str, blob_id: str) -> dict | None:
     """Parses a text/calendar mail attachment (already a blob in the account's JMAP namespace) and
     reports whether its event is on the calendar yet, so the thread view can offer "Add to Calendar".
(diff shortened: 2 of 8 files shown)


s-aga-r:
@greptileai review

PR #556, opened 2026-08-09
refactor(meet): validate frontend protocols

BreadGenie:
- Check API and Socket.IO messages when they enter the Meet frontend.
- Use named types for participants, media, transports, polls, and E2EE data.
- Remove explicit `any` and unsafe `as unknown as` casts from production code.


BreadGenie:
@greptileai review

PR #562, opened 2026-08-09
refactor(meet): use participant connection lifecycle

BreadGenie:
- Replaced the manual Meet setup sequence with one Participant Connection start operation.
- Kept host, guest, waiting-room, auto-mute, media selection, and encryption policy outside the lifecycle Module.
- Moved live lifecycle state out of the page connection store.
- Derived UI loading and setup state from Participant Connection state.
- Hid connection and recovery internals behind meaningful Meet methods.
- Updated recovery diagnostics and tests.


BreadGenie:
@greptileai review

PR #570, opened 2026-08-10
fix(drive): stop content_doctype/content_docname from being forged

ebrahimgamdiwala:
content_doctype/content_docname link a Drive File to the content document it represents, and content_has_permission delegates all permission on that document to whichever File claims the link. Since those fields were plain, freely-editable custom fields, any user could point a File they own at another user's Writer Document (or, via the same has_permission delegation, a Presentation) and inherit full read/write access to it — and deleting that File would cascade-delete the victim's document via after_delete's ignore_permissions=True path.

Restrict content_doctype/content_docname to Drive's own trusted creation paths (create_for_doc, the attachment-reference branch of after_file_upload, and Administrator-run migration patches); block any other insert or update that sets or changes them.


BreadGenie (approved):
LGTM

PR #571, opened 2026-08-10
fix(mail): rank recipient suggestions by relevance

s-aga-r:
## Problem

Searching the email-address index for `doe` could return the person you meant last. The cause was not weak ranking — there was **no ranking**:

1. **Prefix matches come back unscored.** Every hit from Tantivy's `phrase_prefix_query` scores exactly `1.0`, so results were ordered by internal doc id, i.e. *insertion order*. "Jane Doeringer" outranked "John Doe" only because it was indexed first.
2. **Short prefixes silently lost matches.** Phrase-prefix expands a prefix to at most 50 terms. On a 20k-address index, querying `j` returned 20000 hits and excluded `jane@example.com` entirely — one letter made a contact unreachable.
3. **Non-ASCII names could never match.** The query tokenizer was `[a-z0-9]+`, so "Müller" shredded into `m` + `ller` while the index held `müller`.

## Changes

**Retrieval** (`suite/store/search_store.py`) — `search_prefix` joins `search_phrase_prefix`: all terms required, trailing term matched by an automaton-backed prefix query with no expansion cap. Adjacency is no longer required at retrieval, so "jane doe" now also finds "Jane Ann Doe". The docstring states that hits are unranked and ranking is the caller's job.

`search_phrase_prefix` is kept as a deprecated alias with its phrase semantics **intact** — it still runs a phrase-prefix query, so a caller held up by the old name keeps getting the adjacent, in-order match that name promises. It warns and points at `search_prefix`. Deprecating a name is no licence to change what it does.

**Ranking** (`suite/mail/store/indexes/email_address.py`) — matches are scored where the name/local-part/domain structure is known:

```
match quality (name/local over domain, whole, adjacent, position, residual)
  → named, shortest, alphabetical
```

Every match is ranked, not a fixed pool of them. Retrieval is unscored, so the best address can sit anywhere in the match set — cutting the set before the sort would drop it, and whoever was indexed first would win a broad query outright. `search_prefix(limit=None)` fetches the lot: a page of 5000 first, then a refetch sized by the count if that page came up short, so the common query stays one round-trip.

Both passes run against **one searcher**. A searcher holds the segments as they stood when it was made, so the count still describes what the second pass fetches however many writes land meanwhile. Two independent searches would let a write in between leave the fetch short of its own count — and short at the end, which is exactly where a newly indexed address sits.

The schema is unchanged, so no reindex is needed.

## Verification

Built real Tantivy indexes and drove them through the actual classes:

| Check | Result |
|---|---|
| `"doe"` | word-exact match leads, not the longer word that merely starts with it |
| `"jörg"` / `"müller"` | now match (previously impossible) |
| 6001 addresses starting `"jan"`, exact one indexed **last** | leads the suggestions; before the pool fix it was absent entirely |
| a concurrent worker committing an address mid-search | 6000 of 6000 fetched; before the shared-searcher fix, 6000 of 6001, the new address dropped |
| single letter against 20k addresses | ~85ms worst case; ~1ms for anything more selective |
| `search_phrase_prefix("jane d")` over "Jane Doe" / "Jane Ann Doe" / "Doe Jane" | returns "Jane Doe" alone, and warns |
| `search_prefix("jane d")` over the same three | returns all three |

39 unit tests pass, including the reported case as a regression test.

## Known gaps

- **No accent folding** — "muller" still will not find "Müller"; that needs a custom index-time tokenizer plus a rebuild.
- **No frequency signal** — addresses the query matches equally well are separated only by name presence and address length, not by which one you actually write to. That is the follow-up in #568.

### DIFF
--- a/suite/mail/store/indexes/email_address.py
+++ b/suite/mail/store/indexes/email_address.py
@@ -4,10 +4,80 @@ from frappe.utils import EMAIL_MATCH_PATTERN
 
 from suite.store.search_store import FieldSpec, SearchStore
 
-_TOKEN_PATTERN = re.compile(r"[a-z0-9]+")
+# Runs of alphanumerics, Unicode-aware, mirroring the tokenizer the indexed text went through —
+# an ASCII-only pattern would shred "Müller" into "m" + "ller" and never match the indexed "müller".
+_TOKEN_PATTERN = re.compile(r"[^\W_]+")
 
 # Quote characters some clients wrap display names in, e.g. "'Jane Doe'".
 _WRAPPING_QUOTES = "'\"`"
+
+# Candidates pulled from the index before ranking. Tantivy returns prefix matches unscored, i.e. in
+# index order, so the best address can sit anywhere in the match set and the pool has to be deep
+# enough to hold it: this covers an entire address book of that size, and costs ~20ms in the worst
+# case measured — a single-letter query against a 20k-address book, where the next keystroke narrows
+# the field anyway. Anything more selective ranks in well under a millisecond.
+_CANDIDATE_POOL = 5000
+
+# Ranks below every explained match. A hit matched the indexed "<name> <email>" blob, which can span
+# fields — "doe jane" matches "Jane Doe jane@…" — so a candidate need not match any single field.
+_NO_MATCH = (9, 9, 9, 9, 9)
+
+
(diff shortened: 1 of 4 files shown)


s-aga-r:
@greptileai review

PR #574, opened 2026-08-10
feat(mail): rank recipient suggestions by correspondence

s-aga-r:
Recipient suggestions are ranked by how well the query matches a name or address (#571), with no notion of who you actually write to. For `user`, two equally good matches — `user@example.com` and `user@example.org` — are separated only by address length and alphabetical order, so the address you mail daily can sit below one you mailed once.

### What changed

Every indexed address now carries a `count`: sightings past the first, so a new address is inserted at `0` and each later one adds to it. It slots into `_relevance_key` **after** the match-quality tuple and before the named/shortest tie-breaks, so it only ever separates matches the query cannot tell apart — a much-mailed "Jane Doeringer" still ranks below "John Doe" for `doe`, and a busy address at `example.com` never outranks a name match for `exa`.

Accumulating requires an upsert to read the count it replaces, so `SearchStore` grows a `merge_document(document, existing)` hook behind a `MERGE_ON_UPSERT` flag. The read runs under the same write lock as the write — the Redis lock is not reentrant, so a subclass could not take it itself, and reading outside it would lose increments to a concurrent upsert of the same address. Stores that leave the flag off take exactly the old path.

Two counting rules keep the tally meaningful:

- **Every occurrence in a batch counts, not every call.** One batch can carry a page of messages and a rebuild feeds in hundreds, so collapsing them would tally how the caller batched its work rather than how much mail the address is on.
- **Messages count, contact cards do not.** A contact sync re-indexes the whole address book; counting it would lift contacts you never write to above the addresses you do. Contact cards still upsert, refreshing names, with no bump.

### Caching is what marks a message indexed

Only messages the cache has not held before are indexed — otherwise a message re-fetched after a flag change would score sync churn as correspondence. Three things follow from leaning on the cache that way, and each is handled:

**Which messages are new is decided by the write itself.** `DataStore.set_many` returns the keys the store did not already hold, found by the cursor as it writes each one. Asking first and writing second cannot settle it — two fetches racing to cache the same message both look before either commits, both are told it is new, and both count its addresses. Measured with eight concurrent writers on one key, the check-then-write shape told eight of them they had stored it first; deciding inside the write tells exactly one, since LMDB serializes writers. It also drops a transaction from the caching path rather than adding one.

**A message that failed to index does not stay cached.** Otherwise the failure would be permanent — the message would never be offered as new again, and the people on it could be missing from suggestions until someone rebuilt the index by hand. Indexing can fail for ordinary reasons (a write lock not acquired within 30s, a write error), so the batch that did not index is dropped from the cache and picked up on the next fetch, at the cost of one round trip to a server that still has the message. The drop is unconditional even if another request re-cached one of those messages meanwhile: that request skipped indexing on the same grounds, so sparing its copy would leave the message cached with nobody left to index it. Its copy mirrors the server, which flag changes write to first, so dropping it costs a re-fetch rather than state.

**If the store will not give the message up either, a rebuild is queued.** That is the one case nothing local can fix, so it goes to `rebuild_email_address_index`, which reconciles the whole index against the cache and dedupes per account. Both failures are logged, and a repair that cannot even be queued still never reaches the caller.

Adding a field changes the index's schema hash, which wipes each account's index the next time it is opened, so `rebuild_email_address_indexes` is bumped to `#3` to re-run.

### Known, deliberate drift

Counts rank addresses against each other; they are not a ledger. A message evicted and re-fetched counts again, clearing the message cache lets a re-sync count everything a second time, and your own address — on every message you send or receive — accumulates the highest count of all. `rebuild_email_address_index` recomputes counts from the cache whenever they need resetting. Each is documented where someone would trip over it.

### Testing

- 50 unit tests on the index (tally, merge, name preservation, ranking), 18 on the search store (merge hook, chunked lookup, duplicate-ID merge chaining), 13 on message caching (what gets indexed, what gets rolled back, what gets escalated) and 6 on the data store — including a threaded test asserting exactly one of eight racing writers is told a key was new.
- Full mail suite green: 23 modules, 191 tests.
- Against a live index: first insert → `0`, three more sightings → `3`, an uncounted upsert leaves the count and keeps a name a bare sighting would have erased, and `search_email_addresses("user")` returns `user@example.com` (3) ahead of `user@example.org` (1).
- Recovery, with the real store and index: an indexing failure leaves the message uncached and the index empty; the next fetch caches and indexes it at `0`; a later re-cache does not count it again. With the rollback failing too, the message stays cached and unindexed — and the queued rebuild, when run, indexes it.
- 1500 addresses in one call (a real rebuild crosses the 1000-ID chunk boundary): 1200 seen twice land at `2`, the other 300 at `1`, none lost at the boundary.
- Cost at 20k addresses: indexing 40×500 batches 2.75s fresh, 3.71s when every document already exists (~24ms per batch for the merge lookup). Search unchanged — 89ms for the pathological single-letter query, 1.1ms for a selective one.

### DIFF
--- a/suite/mail/doctype/contact_card/contact_card.py
+++ b/suite/mail/doctype/contact_card/contact_card.py
@@ -505,8 +505,13 @@ def _cache_contact_cards(account: str, contact_cards: dict[str, dict]) -> None:
     store.set_many(Entity.CONTACT_CARD, items=contact_cards)
 
     # Feed contact addresses into the shared address index; never let indexing break caching.
+    # Uncounted: the whole address book is re-cached on every contact sync, and a contact card says
+    # who someone is, not how much mail they are on — counting it would rank contacts the user never
+    # writes to above the addresses they actually correspond with.
     try:
-        get_email_address_index(account).index_addresses(_contact_addresses(contact_cards.values()))
+        get_email_address_index(account).index_addresses(
+            _contact_addresses(contact_cards.values()), count=False
+        )
     except Exception:
         log_mail_error(
             _("Failed to index contact addresses for search"), frappe.get_traceback(with_context=True)
--- a/suite/mail/doctype/mail_message/mail_message.py
+++ b/suite/mail/doctype/mail_message/mail_message.py
@@ -1294,11 +1294,12 @@ def _cache_messages(account: str, messages: dict[str, dict]) -> None:
     """
 
     store = get_data_store(account)
-    # Checked before the write, while the store can still tell which of these it has already seen.
-    known = store.exists_many(Entity.EMAIL, keys=list(messages))
-    store.set_many(Entity.EMAIL, items=messages)
+    # Which of these the cache had never held, answered by the write itself: two fetches racing to
+    # cache the same new message would otherwise both be told it was new and count its addresses
+    # twice. Only one of them gets the message back from here.
+    new_ids = store.set_many(Entity.EMAIL, items=messages)
 
-    new_messages = [message for message_id, message in messages.items() if message_id not in known]
+    new_messages = [message for message_id, message in messages.items() if message_id in new_ids]
     if not new_messages:
         return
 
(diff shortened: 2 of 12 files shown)


s-aga-r:
@greptileai review

s-aga-r:
@greptileai review

PR #609, opened 2026-08-12
ci: scope E2E tests to affected apps

BreadGenie:


BreadGenie:
Superseded by #612, which carries the same path-scoping change from a same-repository branch so E2E code does not cross the privileged runner's fork boundary.

PR #625, opened 2026-08-13
fix(mail): realign owner with user on records that predate OwnerFromUser

s-aga-r:
## Problem

A user with only the **Suite User** role gets a 403 when the Mail UI loads the Credentials page:

```
GET /api/method/frappe.client.get?doctype=User+Settings&name=...
[HTTP/2 403]
frappe.exceptions.PermissionError
```

The Suite User role reaches User Settings (and the other `OwnerFromUser` doctypes) through `if_owner`, which keys off the standard `owner` (created-by) field. New records get `owner = user` pinned by the `OwnerFromUser` mixin, but records provisioned before the mixin existed — e.g. User Settings created by the User `after_insert` hook while an admin or the signup flow's Administrator session was active — are still owned by whoever ran the code. The user they belong to fails the `if_owner` check on their own record.

## Fix

A data patch that backfills `owner = user` (wherever they disagree) for all seven doctypes that mix in `OwnerFromUser`: User Settings, User Account, Mail Signature, Mail Queue, Mail Exchange, Contacts Exchange, and Calendar Exchange. One idempotent UPDATE per doctype; `modified` is left untouched.

## Verification

Simulated the broken state locally (Suite User's User Settings with `owner = Administrator`):

- Before the patch: `frappe.has_permission("User Settings", "read", doc=...)` as that user returns False (reproduces the 403).
- After the patch: the failing `frappe.client.get` call succeeds; the user still cannot read other users' settings, and list views return only their own row.
- `bench migrate` runs the patch cleanly.

### DIFF
--- a/suite/patches.txt
+++ b/suite/patches.txt
@@ -100,3 +100,4 @@ execute:frappe.db.set_single_value("Mail Settings", "max_mailing_list_participan
 # Existing, already-set-up sites shouldn't be sent through Suite's onboarding.
 execute:frappe.db.set_single_value("Suite Settings", "is_onboarded", 1) if frappe.is_setup_complete() else None
 suite.suite_core.patches.move_blob_store_out_of_private
+suite.suite_core.patches.align_owner_with_user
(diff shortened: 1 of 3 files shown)


s-aga-r:
@greptileai review

PR #645, opened 2026-08-14
fix(meet): skip recorder-dependent reconciliation when no recorder is configured

s-aga-r:
## Problem

On sites without a recorder server configured, the `reconcile_pending_recordings` scheduled job called `_client()` for every Pending/Stopping recording, raising `ValueError: URL must be configured` and creating an error log per recording on every scheduler run.

## Fix

Guard the two client-dependent phases (querying pending recordings and retrying stops) behind `_fixture_enabled() or _recorder_available()`. When the recorder is not configured these phases are skipped; affected recordings are not left stuck, because the stale sweep in the same job (which needs no client) still fails them with `recorder_unavailable` once past `max_ends_at`, and failed-recording cleanup continues to run.

## Tests

- `suite.meet.api.test.test_recording_reliability` — 12 passed
- `suite.meet.api.test.test_recording` — 25 passed

### DIFF
--- a/suite/meet/api/recording.py
+++ b/suite/meet/api/recording.py
@@ -680,16 +680,20 @@ def recorder_failed(
 
 
 def reconcile_pending_recordings():
-    names = frappe.get_all(
-        "Meet Recording",
-        filters={"status": "Pending", "pending_deadline": ["<=", now_datetime()]},
-        pluck="name",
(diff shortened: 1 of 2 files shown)


BreadGenie (approved):
LGTM

PR #646, opened 2026-08-15
fix: block installing Suite when a standalone suite app is installed

s-aga-r:
## Problem

Installing the Suite app on a site that already has one of the standalone apps it consolidates (Calendar, Drive, Mail, Meet, Sheets, Slides, Writer) leads to conflicting modules and DocTypes.

## Fix

Add a `before_install` hook (dispatched through `suite.suite_core.boot`) that checks the site's installed apps and aborts `install-app suite` with an error listing the conflicting standalone app(s), asking the user to uninstall them first.

Error shown:

> Cannot install Frappe Suite because the following standalone app(s) are installed on this site: **drive**, **mail**. Frappe Suite already includes them. Please uninstall these app(s) first, then install Frappe Suite.

The check runs per-site at install time (bench CLI and Desk install flow), before anything is written.

### DIFF
--- a/suite/hooks.py
+++ b/suite/hooks.py
@@ -327,6 +327,7 @@ scheduler_events = {
 # ============================================================================
 from suite.suite_core import boot as _suite_boot
 
+before_install = "suite.suite_core.boot.before_install"
 after_install = "suite.suite_core.boot.after_install"
 after_migrate = "suite.suite_core.boot.after_migrate"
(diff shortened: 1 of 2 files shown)


BreadGenie:
Should we also let them know how to keep the data as well?
- take backup of the site 
- uninstall individual suite apps
- install suite
- restore backup

s-aga-r:
https://github.com/frappe/suite/pull/654

PR #664, opened 2026-08-17
fix(mail): improve Outbox usability on mobile

s-aga-r:
Fixes several usability issues on the Outbox pages, mostly affecting mobile.

### List view

- **Empty state was clipped on mobile**: the ListView body scrolls horizontally, so the "no submissions" text was centered against the full (off-screen) table width and required scrolling right to read. It now renders outside the ListView, centered in the visible viewport.
- **Pagination → infinite scroll**: the pager was inconsistent with the rest of the user-facing Mail UI. Scrolling near the bottom now appends the next page (deduped by id, since rows can shift while other clients schedule/cancel). The 30s poll, the new-mail socket event, and post-action reloads refetch every loaded page in parallel and swap the rows in place, so scroll position and loaded depth survive background refreshes. Filter/account changes still restart from page 1 behind the skeleton, with a fetch token guarding against stale responses.

### Details view

- **Header actions moved into a dropdown** (desktop and mobile): the row of buttons took nearly the full device width and squeezed the subject out. All actions now live in one `AdaptiveDropdown` (regular dropdown on desktop, bottom sheet on mobile) behind an ellipsis trigger, so long subjects keep the header width and truncate. The action state machine was duplicated between the list rows and this header, so it's extracted into a shared `submissionActions()` builder both pages use.
- **Back button on mobile**: the details page had no way back. `MobileTitleHeader` gains an optional `with-back` prop (styled like the existing folder-sheet hamburger) and the details page uses it to return to the Outbox; desktop keeps the breadcrumb link.

Trade-off: the details page's "Try again now" / "Remove" buttons previously showed a loading spinner; as dropdown items they can't, so feedback comes from the success/error toasts.

All 180 mail frontend unit tests pass.

### DIFF
--- a/frontend/src/apps/mail/components/mobile/MobileTitleHeader.vue
+++ b/frontend/src/apps/mail/components/mobile/MobileTitleHeader.vue
@@ -1,9 +1,9 @@
 <template>
 	<!-- Shared mobile title row (mailbox / all inboxes / screener / profile): 2xl
-	     semibold title, optional xs count, optional folder-sheet hamburger, actions
-	     slot on the right. Without the hamburger the title gets pl-4 (4px row + 16px =
-	     20px) to sit on the px-5 axis of the list content below it; with it,
-	     the button's own inset provides the offset. -->
+	     semibold title, optional xs count, optional folder-sheet hamburger or back
+	     button, actions slot on the right. Without a leading button the title gets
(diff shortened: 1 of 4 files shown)


s-aga-r:
@greptileai review

s-aga-r:
@greptileai review

s-aga-r:
@greptileai review

PR #665, opened 2026-08-17
fix(drive): page lists on the server's `has_next` signal

ebrahimgamdiwala:
## Problem

Drive's list endpoints dedupe and permission-filter **after** applying `LIMIT`/`OFFSET`, so a full SQL window can return fewer visible rows once filtered. The client inferred "end of list" from `page.length >= PAGE_SIZE`, so one filtered row permanently killed infinite scroll — and with no "load more" button, the rest of the folder became unreachable. Worst in Shared, where the dedupe fires on every multi-path share.

## Fix

The five paginated endpoints take an opt-in `paginated` flag and return `{rows, has_next, next_start}`. The paginated path walks raw SQL windows until it has a full page of rows the caller can actually **see**, and derives `has_next` from that — never from the raw count, which would make an empty page advertise the existence of denied files. Filling a page can span several windows, so the response carries `next_start` for the client to resume from. Non-paging callers keep the bare list, untouched.

Three further fixes this turned up:

- **Cache corruption on revisit** — frappe-ui's offline restore assigns the raw response into `resource.data`, where consumers spread it as an array. The new envelope threw a `TypeError` on the second visit to any list.
- **Infinite scroll never fired on a cold first visit** (pre-existing) — `useInfiniteScroll` binds its target once at setup, but the shell registers the scroll container a tick later, so it caught `null` and never re-armed.
- **Stale pagination race** — a refresh while `loadMore` was in flight appended rows from the old query and overwrote the reset cursor. Requests now carry a query epoch.

Also: `loadMore` skipped the dotfile filter page 1 applies, so dotfiles appeared from page 2 onward.

## Testing

7 backend, 11 frontend — each regression case verified failing without its fix. Verified in a browser on a built site: a cold visit fills a full 50-row page past denied rows and reaches all 55 visible files. All Drive backend modules and the full frontend suite pass.


BreadGenie:
I've added few fixes on top

- Prevent hidden-file leaks after the scan limit.
- Deduplicate files shared through multiple permission paths.
- Match backend pagination order with the client’s effective modified value, preventing files from appearing above after scrolling.
- Continue loading while still near the bottom.
- Ignore pagination responses after navigation.

PR #666, opened 2026-08-17
feat(mail): show quota usage in admin dashboard members list

s-aga-r:
## Summary

Shows each member's storage quota usage in the Mail Admin Dashboard → Members list, so admins can spot accounts running out of space without opening each member's page.

- **Backend**: `get_members` now attaches a `quota` object per row. Users are mapped to their personal JMAP account, then all Stalwart accounts' `quotas`/`usedDiskQuota` are read in a single bulk call (no per-member round trips), normalized via the existing `_build_quota_usage` helper. If a member has no personal account, or Stalwart is unreachable, `quota` is `null` and the list still loads.
- **Frontend**: new **Storage** column with a compact usage bar + percentage (fill turns red above 80% used, matching the sidebar `QuotaBar` convention), `∞` for unlimited accounts, and `—` when there's no quota data. Exact figures ("1.2 GB of 5 GB used") are shown in a tooltip on hover.

### DIFF
--- a/frontend/src/apps/mail/pages/dashboard/UsersView.vue
+++ b/frontend/src/apps/mail/pages/dashboard/UsersView.vue
@@ -64,6 +64,28 @@
 								:theme="row.enabled ? 'green' : 'gray'"
 							/>
 						</template>
+						<template v-else-if="column.key === 'storage'">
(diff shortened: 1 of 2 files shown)


s-aga-r:
@greptileai review

PR #705, opened 2026-08-22
fix(slides): save slides by client_id with conflict detection

s-aga-r:
## Problem

A presentation's `Slide` rows could be wiped when two editors had the same deck open (e.g. a presenter's laptop plus the author's).

1. Slides added/pasted/duplicated in the editor carried `name: ''`, and the store never adopted the server-assigned name after a save. Frappe therefore treated them as new rows on **every autosave**: delete + re-insert under fresh names.
2. Frappe's `update_child_table` deletes every row not in the saved list, then runs `UPDATE … WHERE name=` for named rows — a stale name matches 0 rows, silently.
3. A second editor saving its snapshot (autosave, or the unconditional `saveCurrentState()` on unmount/reconnect) deleted the author's current rows and "updated" names that no longer existed → empty deck.

## Steps to reproduce (on `develop`)

Two browser sessions with **write** access to the same deck — two browsers/profiles on one machine, same account is fine.

1. **Browser A**: open a presentation in the editor. Add 2–3 slides (new, duplicate or paste). Wait for autosave. Keep the tab open.
2. **Browser B**: open the same presentation in the editor. It now holds the current row names. Keep it open.
3. **Browser A**: make any edit. Autosave re-inserts the slides from step 1 under new row names (they still carry `name: ''` in A's memory).
4. **Browser B**: make any tiny edit, or just navigate away from the editor — both pushed B's stale snapshot.
5. Reload the deck in either browser, or open `/desk/presentation/<name>`: the Slides table is empty.

The more slides were created in A's session without a reload, the more is lost; a freshly combined deck (every slide added that session) ends up completely empty.

Scripted equivalent (`apps/suite/suite/tmp_repro.py`, run with `bench --site <site> execute suite.tmp_repro.run`):

```python
import frappe
from frappe.client import set_value
from suite.slides.doctype.presentation.presentation import create_presentation

def rows(name):
    return frappe.get_all("Slide", {"parent": name, "parenttype": "Presentation"}, pluck="name")

def run():
    frappe.set_user("Administrator")
    tpl = frappe.get_all("Presentation", {"is_template": 1}, pluck="name")[0]
    p = create_presentation(template=tpl)
    # what the editor sends for slides it created: name ''
    unnamed = [{"name": "", "elements": "[]", "background": "#fff", "client_id": f"c{i}"} for i in range(3)]

    set_value("Presentation", p.name, {"slides": unnamed})
    print("after user's save #1: ", rows(p.name))
    # second client loads the deck now and holds these names
    presenter_copy = [s.as_dict() for s in frappe.get_doc("Presentation", p.name).slides]
    set_value("Presentation", p.name, {"slides": unnamed})
    print("after user's save #2: ", rows(p.name))   # same slides, new names
    set_value("Presentation", p.name, {"slides": presenter_copy})
    print("after presenter's save:", rows(p.name))  # []
    frappe.db.rollback()
```

```
after user's save #1:  ['7b4e1clnuu', '7b4esihrdc', '7b4offmksi']
after user's save #2:  ['7b42b1f4nv', '7b4ib7r7c5', '7b4ve5bggl']
after presenter's save: []
```

With this branch, step 4 shows "This presentation was changed elsewhere. Showing the latest version." in Browser B and the deck stays intact; the scripted case is covered by `test_stale_snapshot_is_refused_and_changes_nothing` and `test_rows_keep_their_name_across_saves`.

## Fix

- New `suite.slides.api.slides.save_slides(name, slides, base_modified)` replaces `frappe.client.set_value` for slide saves. Rows are matched on `client_id` (the identity the editor owns), so a slide keeps one row across saves. Only Slide value fields are accepted; `name`/`parent`/`idx`/`owner` from the client are ignored.
- Snapshots built on an older `modified` than the server's are refused with `TimestampMismatchError`, so a stale tab can no longer delete another editor's rows.
- Editor: sends `presentationDoc.modified` as the base, adopts the returned version, and on conflict marks the local draft clean and reloads the server version with a toast. Unmount and reconnect now save only when dirty.
- Removed the `slide.name = ''` bookkeeping; the payload no longer carries row names.

## Tests

- `suite/slides/api/test_slides.py`: rows keep names across saves, follow the editor list, duplicate `client_id` keeps both slides, stale snapshot is refused without changing rows, framework fields not editable, write permission required.
- `saving.test.ts`: a refused snapshot is dropped and triggers a reload.
- Removed two `commands.test.ts` cases that asserted the old name-blanking.

## Behaviour note

On conflict the losing editor's unsynced edits are dropped in favour of the server's version (with a toast). A `client_id`-keyed merge can be added later on the same API.

### DIFF
--- a/frontend/src/apps/slides/SlidesShell.vue
+++ b/frontend/src/apps/slides/SlidesShell.vue
@@ -13,7 +13,7 @@ import { h, onMounted, onUnmounted, provide, ref, watch } from 'vue'
 import { onBeforeRouteLeave } from 'vue-router'
 import { toast, FrappeUIProvider } from 'frappe-ui'
 import { Wifi, WifiOff } from 'lucide-vue-next'
-import { saveCurrentState } from '@/apps/slides/stores/saving'
+import { saveChanges } from '@/apps/slides/stores/saving'
 import { inSlideShowMode } from '@/apps/slides/stores/slideshow'
 import { setupTheme } from '@/utils/setupTheme'
 import { postToServiceWorker } from '@/apps/slides/utils/serviceWorker'
@@ -32,7 +32,7 @@ const handleOffline = () => {
 
 const handleOnline = () => {
   isOnline.value = true
-  saveCurrentState()
+  saveChanges()
   if (inSlideShowMode.value) return
   toast('You are back online.', {
     icon: () => h(Wifi, { class: 'size-4' }),
--- a/frontend/src/apps/slides/pages/PresentationEditor.vue
+++ b/frontend/src/apps/slides/pages/PresentationEditor.vue
@@ -109,7 +109,7 @@ import {
 } from '@/apps/slides/stores/historyMeta'
 
 import { useShortcuts, showShortcutsModal } from '@/apps/slides/composables/useShortcuts'
-import { saveChanges, saveCurrentState, dirty } from '@/apps/slides/stores/saving'
+import { saveChanges, dirty } from '@/apps/slides/stores/saving'
 import {
 	refreshOfflineStatus,
 	warmOfflineCopyAssets,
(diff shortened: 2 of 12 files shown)


s-aga-r:
@greptileai review

GursheenK (approved):

PR #710, opened 2026-08-23
fix(calendar): don't keep sender copies of custom invite emails

s-aga-r:
## Summary
- Custom invite/update/cancel emails and the attendee's iTIP REPLY were moved to the sender's Sent mailbox after submission, so every invite blast piled up in the organizer's account.
- Pass `destroy_after_submit=True` on both `MailQueue._create` calls in `invitations.py`, so the JMAP `EmailSubmission/set` uses `onSuccessDestroyEmail`: recipients get the email, the sender's copy is destroyed. The calendar event itself remains the record.
- `notify_organizer_of_response` (HTTP-RSVP heads-up) is unchanged — it sends via `frappe.sendmail` from the site account and never landed in a user's Sent folder.

## Test plan
- `bench --site s2 run-tests --module suite.calendar.tests.test_calendar_mail_invites` passes (2/2).
- Manual: with "custom event invites" enabled in Mail Settings, create an event with guests → guests receive the invite with `.ics`; organizer's Sent folder has no copy. Same for an attendee RSVP reply.

### DIFF
--- a/suite/calendar/doctype/calendar_event/invitations.py
+++ b/suite/calendar/doctype/calendar_event/invitations.py
@@ -239,6 +239,9 @@ def notify_organizer_of_reply(account: str, event_id: str, responder_email: str,
             raw_message=message,
             via_api=True,
             delivery_mode="Enqueue",
+            # The reply is a mechanical iTIP message, not correspondence — don't leave a copy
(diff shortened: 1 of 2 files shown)


s-aga-r:
@greptileai review

PR #715, opened 2026-08-24
fix(drive): stop querying dropped old_name field in translate_old_name

s-aga-r:
## Summary
- `translate_old_name` still filtered `File` on `old_name`, a Drive File field that was dropped when Drive File merged into the framework `File` doctype — every legacy `/drive/{file,folder,document}/<id>` link (guest-reachable) threw `(1054, "Unknown column 'old_name' in 'WHERE'")` into the Error Log.
- The old-id mapping data no longer exists on migrated sites, so the endpoint now does the only translation still possible: pass through ids that survived migration as `File` names (`frappe.db.exists`), returning `None` for lost pre-restructure ids.
- The three legacy redirect routes in the frontend fall back to the original id when translation returns nothing, so the target page renders its normal not-found error state instead of navigating with a `null` param.

## Test plan
- Hit `/api/method/suite.drive.api.files.translate_old_name?old_name=<current File name>` → returns the name.
- Hit it with a stale pre-restructure uuid (e.g. `38ea275a07614e63a0fdb90214bf90ed`) → returns `null`, no Error Log entry.
- Visit `/drive/file/<stale-id>` as Guest → lands on the file page's error state instead of a 500.

### DIFF
--- a/frontend/src/apps/drive/routes.ts
+++ b/frontend/src/apps/drive/routes.ts
@@ -154,7 +154,8 @@ export const routes: RouteRecordRaw[] = [
           return {
             name: 'drive-Folder',
             params: {
-              entityName: translate.data,
+              // Untranslatable ids fall through so the page shows its error state
+              entityName: translate.data || to.params.entityName,
             },
           }
         },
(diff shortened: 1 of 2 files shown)


s-aga-r:
@greptileai review

PR #716, opened 2026-08-24
mail: thread view cleanup

krantheman:
Cleans up the thread reading view. What made it feel busy was repetition rather than volume — every collapsed message repeated the recipient roster, the sender's raw address, and stated time twice.

**Collapsed messages are one line** — avatar · sender · preview · clock time. Roster and raw address are gone from them; a collapsed row costs ~48px instead of ~100px, so a real eight-message thread fits on screen.

**Expanded header** shows "to Pushkar, Rushabh and 2 others" beside the sender, chevron next to it for the full list.

**Details popover redesigned** — sender header (avatar, name, address) over a lowercase to / cc / bcc / reply to / subject / date grid. Reply-To only shows when replies actually leave the sender. Long addresses wrap.

**Thread tightened** — slimmer collapsed padding, smaller avatars, closer cards, borderless quieter date dividers. Rows show clock time when day dividers carry the date. Name, preview and time share one baseline. Row actions (star / reply / ⋯) appear on hover at a fixed right edge, timestamp yields, row height never changes.

**Quote toggle labelled** — the bare "···" chip is now a subtle-style button saying "Show/Hide trimmed content", finger-sized on mobile.

**Fix**: the reading pane forced `height:auto !important` on every email image, so an explicitly sized 40×40 avatar with `border-radius:50%` rendered as an ellipse whenever the source photo wasn't square. Images now render as authored; only images the width clamp actually shrinks shed their height.

### DIFF
--- a/frontend/src/apps/mail/components/EmailContent.vue
+++ b/frontend/src/apps/mail/components/EmailContent.vue
@@ -159,15 +159,26 @@ const collapseQuotes = (doc: Document) => {
 		// Only the outermost quote gets a toggle — hiding it hides any quotes nested inside.
 		if (quote.parentElement?.closest('.gmail_quote, .frappe_mail_quote')) return
 		quote.classList.add('quote-hidden')
+		// A labelled control, not a bare '···' chip — unlabelled, it was easy to miss
+		// that a reply hides a whole conversation underneath it.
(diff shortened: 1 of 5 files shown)


krantheman:
@greptile review

PR #727, opened 2026-08-25
fix(mail): self-heal push subscriptions and seed missing sync state

krantheman:
Mail realtime rides on JMAP push webhooks: they trigger `fetch_changes`, which publishes the `new_mail_created` socket event the views refresh on. Without webhooks, no realtime event is ever published and new mail only surfaces through the 30s poll or a manual refresh.

Two bugs made webhooks easy to lose and impossible to recover:

**Subscriptions were never created, only renewed.** No code path called `add_push_subscription`, and the daily job only renews subscriptions that already exist. A subscription lost to a failed creation, an unrenewed expiry or a server-side delete silently ended the user's webhooks forever. `ensure_push_subscription` now looks up this site's subscription on the mail server (the device client id is deterministic per site+user) and creates one when it's missing or expired — wired into the daily job and an enqueued `on_login` hook.

**A missing sync state could never initialize itself.** `fetch_changes` with no stored state and no `email_state` argument (any manual or scheduled run) stored `None`, so every subsequent run just "initialized" again and changes were never fetched. It now seeds from the server's actual Email state via a new `CoreService.get_state()`.

Verified end to end against a live mail server (subscription creation → PushVerification → StateChange webhooks → socket refresh in the browser). Unit tests cover the seeding paths and the healing decision table.

### DIFF
--- a/suite/hooks.py
+++ b/suite/hooks.py
@@ -276,6 +276,10 @@ user_invitation = {
 # Suite's onboarding replaces the built-in desk setup wizard
 setup_wizard_url = "/suite/setup"
 
+# Heal the user's JMAP push subscription on login (enqueued; a lost subscription silently
+# ends webhooks — realtime events and mailbox-count invalidation both ride on them)
+on_login = ["suite.mail.doctype.push_subscription.push_subscription.on_login"]
+
 # ============================================================================
 # Scheduled Tasks (per-frequency lists combined; cron keys de-duplicated)
 # ============================================================================
--- a/suite/mail/doctype/mail_message/mail_message.py
+++ b/suite/mail/doctype/mail_message/mail_message.py
@@ -1407,6 +1407,25 @@ def fetch_changes(user: str, account: str, email_state: str | None = None, ctx: 
     ctx["email_state"] = email_state
 
     if not current_state:
+        if not email_state:
+            # A manual or scheduled run carries no state, and storing None would leave the
+            # account re-"initializing" on every run with changes never fetched — seed from
+            # the server's actual Email state instead.
+            try:
+                email_state = EmailService(account, get_jmap_connection(user)).get_state()
+                ctx["email_state"] = email_state
+            except Exception:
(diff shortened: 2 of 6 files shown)


s-aga-r:
Pushed a series of review fixes on top of the original healing commit. Summary for reviewers:

- **33d1e37** serializes `ensure_push_subscription` under a per-user, site-scoped `filelock`, so login healing and the daily renewal job cannot both observe the subscription as missing and create it twice. A run that cannot get the lock skips quietly, since the holder is doing the same work.
- **ecd0a83 / ee15998** decorator corrections: `get_site_device_client_id` is only called server-side, so it is no longer whitelisted. That removal also dropped the `@frappe.whitelist()` that had originally belonged to `add_push_subscription` (the new functions were inserted between the decorator and the def), which broke the Add Push Subscription modal's RPC; it is restored, and the whitelisted set now matches develop exactly.
- **e7f2549** extracts a `_heal_push_subscription` core that also deletes this site's expired subscriptions (best-effort: a failed delete never blocks creating the replacement), and the daily job now does a single `PushSubscription/get` per user, shared by healing and the expiry scan. Ids healing just deleted are excluded from renewal.
- **210c1c0** dedupes login healing jobs per user (`job_id`, `deduplicate=True`), so a burst of logins from multiple tabs or devices queues one job instead of one apiece.
- **e2ec2d0** tightens the renewal scan to `now < expires <= cutoff`: renewing an already-expired subscription cannot revive it, and attempting it produced daily notUpdated error logs for dead foreign device subscriptions until the server purged them.
- **406da8d** adds direct unit tests for `CoreService.get_state` (empty-get request shape, state unwrap from the first method response, None for error or missing responses).

Tests: 16 healing/renewal cases plus 12 push-sync cases, all green locally; pre-commit clean.

s-aga-r:
@greptileai review

s-aga-r:
@greptileai review

s-aga-r:
@greptileai review

s-aga-r (approved):

PR #728, opened 2026-08-25
fix(mail): preserve the original's formatting when forwarding

krantheman:
Forwarding a rich email (a notification, a marketing mail, a Google Docs share) destroyed its formatting: the original was placed into `html_body`, i.e. into the compose editor, whose schema re-serializes the HTML — unknown tags/attributes dropped, inline styles stripped, tables rebuilt as bare editor tables. Replies never had this problem because their quote lives in `quoted_content`, outside the editor, and is concatenated back verbatim at send time.

Forward now uses the same mechanism: the forwarded original (with its From/Date/Subject/To header block) goes to `quoted_content`, and `extractQuotedContent` additionally recognizes the `frappe_mail_fwd` wrapper so a saved forward draft re-splits on reopen instead of feeding the original to the editor. In the composer the forwarded content sits behind the existing quoted-content toggle; expanding it (and thereby editing it) is what hands it to the editor — the user's explicit choice, as with replies today.

Preserving the original verbatim exposed a second leak: an email is authored as a full document, and its `body`/`html` style rules repainted the *entire* outgoing mail — e.g. a dark marketing backdrop bleeding over the sender's own text. The new `containEmailHtml` keeps the original's styles but scopes every selector to the embedded block (`body`/`html`/`:root` become the container itself, other selectors are prefixed; the body's `bgcolor`/`style` moves onto the container; external stylesheets are dropped). Any parse failure falls back to the raw HTML so containment can degrade but content can't be lost. Replies get the same containment.

Verified by forwarding an AWS Health notification and a Google Docs share mail: branding, buttons and tables arrive intact, with the original's backdrop confined to the forwarded block.

---

Also rides along: two small thread-view refinements — the quoted-content toggle sizes by the app's own mobile state instead of the iframe's media query, and the message-details popover shows the sender as a "from" row in place of the avatar header.

### DIFF
--- a/frontend/src/apps/mail/components/EmailContent.vue
+++ b/frontend/src/apps/mail/components/EmailContent.vue
@@ -65,7 +65,7 @@ import { Button } from 'frappe-ui'
 
 import { analyzeRemoteAssets, blockRemoteAssets } from '@/apps/mail/utils'
 import { escapeBracketedAddresses } from '@/apps/mail/utils/html'
-import { useComposeMail, useTheme } from '@/apps/mail/utils/composables'
+import { useComposeMail, useScreenSize, useTheme } from '@/apps/mail/utils/composables'
 import { parseMailto } from '@/apps/mail/utils/mailto'
 import {
 	declaresFixedPalette,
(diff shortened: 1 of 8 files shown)


krantheman:
@greptileai review

PR #734, opened 2026-08-25
feat(meet): apply dynamic rate limits to all public endpoints

### LINKED ISSUE #721: [Improvement]: Missing Rate Limiting on public-facing Meeting endpoints
### App

Meet

### What happened?

Public-facing endpoints for joining meetings and approving users do not have any rate limits applied. This leaves the backend and the host's client vulnerable to spam and minor DoS attacks

1. join_meeting_as_guest 
2. approve_user

### Steps to reproduce

na

### Environment

Suite App Meet Module (latest), Self-hosted
MacOS
Brave

dankgarlic1:
This PR migrates all public and guest-facing meeting endpoints to use the @dynamic_rate_limit decorator.

It replaces the hardcoded static limits for existing endpoints and adds protection to the previously unprotected ones (like joining and approvals) to prevent spam and minor DoS attacks. I also included an install.py script and a patch so existing and new installations will automatically get the database records for these rate limits seeded with sensible defaults.

Endpoints updated:

- create
- get_public_meeting_preview
- join_meeting
- approve_join_request
- approve_all_join_requests
- reject_join_request
- join_meeting_as_guest
- get_approved_guest_connection_details
- get_guest_sfu_connection_details
- validate_guest_session
- check_meeting_access

closes #721


suite/meet/install.py
```diff
@@ -0,0 +1,47 @@
...
+        {"method_path": "suite.meet.api.meeting.get_public_meeting_preview", "limit": 10, "seconds": 60},
+        {"method_path": "suite.meet.api.meeting.join_meeting_as_guest", "limit": 10, "seconds": 60 * 60},
+        {
+            "method_path": "suite.meet.api.meeting.get_approved_guest_connection_details",
+            "limit": 10,
+            "seconds": 60 * 60,
+        },
+        {"method_path": "suite.meet.api.meeting.validate_guest_session", "limit": 10, "seconds": 60 * 60},
+        {"method_path": "suite.meet.api.meeting.check_meeting_access", "limit": 10, "seconds": 5 * 60},
+        {"method_path": "suite.meet.api.meeting.join_meeting", "limit": 30, "seconds": 60, "ip_based": False},
```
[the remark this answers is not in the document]
dankgarlic1: need your opinion @BreadGenie
If we rate limit meeting joins by IP address, we are effectively rate limiting an entire network instead of individual users.

For example, if 50 students in a classroom or 50 employees in an office join a meeting at the same time, they may all appear to come from the same public IP address because they are using the same network. With a limit of 30 requests per minute, the 31st legitimate user would get blocked.

This is basically the "noisy neighbor" problem. I found a good explanation of it https://zuplo.com/blog/dont-rate-limit-by-ip

**What I changed**

To fix the CI validation issue and also handle the shared network problem, I updated `suite/meet/install.py`.

*Guests (`get_guest_sfu_connection_details`)*

I set `ip_based` to `False` and added `key: "guest_token"`.

This gives each guest their own rate limit bucket, since Frappe can get the guest token directly from the request payload. So one guest cannot affect another guest just because they are on the same network.

*Authenticated users (`join_meeting`)* 

This endpoint only receives `meeting_id` in the payload. Frappe's rate limiter currently expects the rate limit key to come from the request arguments, so we cannot easily use `frappe.session.user` here and still satisfy the database validation.

As a temporary compromise, I am leaning towards keeping this IP based but increasing the limit to 60 requests per minute.

I know that increasing the limit is not really a proper fix. It is basically just giving the shared network more headroom, while also making the rate limit a bit less effective. So I want to be clear that I see this only as a temporary workaround, not the actual solution.

this is what i wanna do next

I think the proper fix should be done at the architecture level.

We could add a small wrapper for Suite's rate limited endpoints that automatically uses `frappe.session.user` when the request is from an authenticated user. This would let us rate limit authenticated users based on their actual user identity instead of their IP address.

It would also mean that we do not have to depend on request payload arguments for the rate limit key. More importantly, it would solve the shared network problem properly instead of just increasing the limit.

i read this for ref: https://owasp.org/API-Security/editions/2023/en/0xa4-unrestricted-resource-consumption/

I would rather open a follow up issue for this and keep this PR focused instead of making the current change more complicated.

s-aga-r:
```suggestion
        {"method_path": "suite.meet.api.meeting.join_meeting", "limit": 30, "seconds": 60, "ip_based": 0},
```

suite/meet/install.py
```diff
@@ -0,0 +1,47 @@
...
+        {
+            "method_path": "suite.meet.api.meeting.get_guest_sfu_connection_details",
+            "limit": 30,
+            "seconds": 60,
+            "ip_based": False,
```
s-aga-r:
```suggestion
            "ip_based": 0,
```

s-aga-r (changes requested):

suite/meet/install.py
dankgarlic1: will address this also along with the ip issue

suite/meet/install.py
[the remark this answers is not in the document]
dankgarlic1: @BreadGenie hello please let me know when you can take a look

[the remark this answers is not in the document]
BreadGenie: Sounds good. Let's not expand the scope of this PR. Go ahead with your implementation.
cc @s-aga-r

dankgarlic1:
@BreadGenie this can be reviewed

suite/meet/api/test/test_meeting.py
```diff
@@ -360,7 +360,7 @@ def test_guest_proof_resumes_same_room_principal(self):
     def test_approved_guest_connection_details_is_post_only(self):
         self.assertEqual(
             frappe.allowed_http_methods_for_whitelisted_func[get_approved_guest_connection_details],
-            ("POST",),
+            ["POST"],
         )
```
dankgarlic1: hello I actually changed the test because I thought this musut have been left behind from the refactor you did, because its comparing exactly ["POST",] but in the meeting.py file after I rebased I saw we used methods=["POST"] only hence I refactored the test instead of adding a trailing comma to every method 

pls lmk what you think @BreadGenie

BreadGenie: You can revert this change. Framework internally converts it to a tuple.

BreadGenie: On 2nd thought wth is that test lol. Checking if it's a post request. I think we can just remove it.

dankgarlic1: @BreadGenie done, if everything else looks good we can merge this?

BreadGenie (approved):
LGTM
@s-aga-r can you check too?

BreadGenie:
rebased to get e2e tests running, but ig I borked the history 😬

BreadGenie:
umm yeah this was the reason I skipped using github runners

s-aga-r:
@BreadGenie LGTM

s-aga-r (approved):

PR #736, opened 2026-08-26
feat(calendar): include the meet link in the ICS LOCATION

s-aga-r:
## Summary
- Append the Frappe Meet link to the `LOCATION` property in generated ICS (invite emails and calendar export), matching Google Calendar's `LOCATION:Remote; https://meet.google.com/...` behavior, since many clients only surface `LOCATION` while `CONFERENCE`/`URL` get buried.
- When the event has no physical location, `LOCATION` carries just the meet link; when the location text already contains the link, it is not appended twice.
- `CONFERENCE` and `URL` properties are unchanged and still emitted alongside.

## Test plan
- Built REQUEST ICS via `build_event_ics` for three cases: location + meet link (`LOCATION:Remote\; https://.../meet/...`), meet link only (`LOCATION:https://.../meet/...`), and location text already containing the link (no duplication).


s-aga-r:
@greptileai review

PR #757, opened 2026-08-30
refactor(meet): extract local capture session

BreadGenie:
## Summary

- extract camera and microphone lifecycle state into `LocalCaptureSession`
- stage capture and publication changes before committing local media state
- isolate cancellation, owner handoff, and E2EE partial-failure handling from Vue

Depends on #756.


BreadGenie:
@greptileai review

PR #769, opened 2026-08-31
feat(mail): make the undo send period configurable per user

s-aga-r:
## Summary
- `User Settings` gets a Compose section with `undo_send_period` (Select: 5, 10, 20, 30 seconds, default 5, required). Existing rows pick up the column default, so no patch is needed.
- The server no longer holds a plain Send for a fixed 10 s + 3 s: `get_undo_send_hold_seconds()` uses the sender's period plus the 3 s grace, with a fallback to the default when the row is missing or holds an off-list value. `get_user_info` exposes `undo_send_period`.
- The composer's Undo toast lives for the user's period, read per send so a change in Settings applies to the next send without reopening the composer.
- New Compose tab in Mail settings (desktop dialog and mobile list, after Signatures) with an "Undo Send" row to edit the period; shared periods/default live in `utils/undoSend.ts`.

## Test plan
- `bench --site <site> run-tests --module suite.mail.doctype.user_settings.test_user_settings` (fallback + Select validation)
- `bench --site <site> run-tests --module suite.mail.tests.test_mail_scheduled_send --test test_undo_send_holds_and_cancels --test test_undo_send_hold_follows_user_settings` (needs Stalwart; the second sets 30 s and checks the hold grows)
- Manually: Settings > Compose > pick 20 seconds > Save; send a mail and confirm the Undo toast outlives 10 s and Undo still returns the message to Drafts. Verified against a live Stalwart: hold was ~22 s server-side and the toast was still up at 14 s.

### DIFF
--- a/frontend/src/apps/mail/components/Modals/SettingsModal.vue
+++ b/frontend/src/apps/mail/components/Modals/SettingsModal.vue
@@ -36,6 +36,7 @@ import {
 	KeyRound,
 	Mailbox,
 	Palette,
+	PenLine,
 	TreePalm,
 	User,
 	Zap,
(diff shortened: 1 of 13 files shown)


s-aga-r:
@greptileai review

PR #774, opened 2026-08-31
feat(mail): lay the submission details page out as a ledger

s-aga-r:
## Summary
- The Outbox's submission details page is one narrow column instead of four identical cards: the subject, the merged delivery state ("Retrying", not the raw undoStatus "Final") with a one-clause summary, the state's actions inline, then an activity list in time order, and the message / envelope / identifier facts as flat sections. Empty rows are gone and the sender's address appears once.
- `utils/submissionActivity.ts` derives the summary clause and the activity entries from the details: the release, each recipient's standing with its raw SMTP reply, and the next attempt. Everything comes from the JMAP submission and the MTA queue, whichever client sent the mail, so only the release and the next attempt carry a time.
- New `SubmissionActivity`, `LedgerSection` and `LedgerRow` components; `priorityLabel` moved into `utils/submission.ts`.
- The Outbox actions no longer write to the Mail Queue: `_sync_queue_log` mirrored the replacement submission id and the send / cancel moments into the row that logged the original submission, for the desk form only. Nothing read those fields, and overwriting `submitted_at` skewed `submitted_after`; the row stays a log of what was submitted, and the server's EmailSubmission is the only state that matters.

## Test plan
- `npx vitest run src/apps/mail`: 239 pass, including 10 new cases for `statusSummary` / `activityEntries` across scheduled, sent, retrying, failed, delivered and cancelled sends.
- `bench run-tests --module suite.mail.tests.test_mail_scheduled_send`: 23 pass (the assertions on the mirrored Mail Queue fields are gone; the ones on the original submit remain).
- Built and checked on a bench against real submissions: a scheduled send (send now / reschedule / cancel inline), a sent one, a cancelled one and a deleted-message one, on desktop and phone (the last section clears the tab bar and compose button).
- The actions (send now, reschedule, cancel, try again now, send again, remove) are the unchanged resources; replacements still follow to the successor's page.

### DIFF
--- a/frontend/src/apps/mail/components/LedgerRow.vue
+++ b/frontend/src/apps/mail/components/LedgerRow.vue
@@ -1,0 +1,12 @@
+<template>
+	<div class="flex items-baseline gap-4 py-1.5">
+		<div class="text-ink-gray-5 w-20 shrink-0 text-sm sm:w-36">{{ label }}</div>
+		<div class="text-ink-gray-9 min-w-0 flex-1 text-base leading-[18px] break-words">
+			<slot>{{ value || '—' }}</slot>
+		</div>
+	</div>
+</template>
+
+<script setup lang="ts">
+defineProps<{ label: string; value?: string }>()
+</script>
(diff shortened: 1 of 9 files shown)


s-aga-r:
@greptileai review

PR #775, opened 2026-08-31
feat(calendar): event reminders that reach the browser

krantheman:
Reminders existed as data — the server fires a `CalendarAlert` at the trigger time — but the only delivery was an FCM device push, and only where push notifications were configured. Now:

- The alert handler publishes a `calendar_alert` realtime event to the user before attempting the push, so any open suite tab hears it.
- In the browser it becomes a toast with an **Open** action *and* a system notification (when permitted): the OS handles Do Not Disturb and stacking, the `tag` dedupes across tabs. Clicking either lands on the event.
- It works from any app with a socket — calendar, mail, meet, drive — via a new shared `createSiteSocket()` that replaces the four drifting copies of the same URL arithmetic. The never-published `refetch_resource` listener is removed.
- The Alert picker offers what actually exists: **Notification** and **Email**. "Sound Alert" is gone (an imported `Audio` alert reads as Notification), and adding a first reminder is when the browser asks for notification permission.

Verified end-to-end against the real mail server (webhook → job → toast + notification + phone push). Two known gaps, not in this PR: the server does send `email` alarm mails (verified on the minute), though it appears to skip them when a display alert shares the same trigger time — upstream repro being narrowed, The invitee gap is closed in this PR too: calendars are seeded with default alerts (10 min before timed events, 9 am for all-day), invite copies use them, and an event saved with no explicit reminder falls back to them — verified end to end, two seeded events alarmed to the second.

### DIFF
--- a/frontend/src/apps/calendar/components/AppSidebar.vue
+++ b/frontend/src/apps/calendar/components/AppSidebar.vue
@@ -72,7 +72,9 @@ const upcoming = computed(() => {
 			// An all-day event covers the whole of today; a timed one is over once its end has passed.
 			return event.isAllDay || dayjs(`${event.toDate} ${event.toTime}`).isAfter(current)
 		})
-		.sort((a, b) => a.fromDateTime.localeCompare(b.fromDateTime))
+		// Sorted on the shape transformEvent hands over — date plus wall clock. An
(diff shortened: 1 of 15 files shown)


krantheman:
@greptileai review

PR #776, opened 2026-08-31
feat(calendar): pre-filled default reminders, all-day alert fixes

krantheman:
New events open with the default reminder pre-filled as an ordinary, editable alert row — 10 minutes before a timed event, 9am on the day of an all-day one — instead of an invisible empty-means-defaults rule. Removing the row now really means no reminder. Flipping All Day swaps an untouched row between the two defaults, and an untouched dated row follows the event when its start date moves.

Fixes along the way:

- The modal never sent `show_without_time`, so all-day events were stored as timed midnight events and got the timed default alert (10 minutes before midnight).
- A server alert without `relativeTo` rendered an empty relative-to select; it now defaults to start, per the spec.

### DIFF
--- a/frontend/src/apps/calendar/components/Modals/EventModal.vue
+++ b/frontend/src/apps/calendar/components/Modals/EventModal.vue
@@ -592,11 +592,18 @@ const shouldShowRecurringEventModal = computed(
 
 // --- Alerts ---
 
+// Touching the alert UI is a choice about reminders: from here on an empty
+// list means cleared, not "still following the calendar's defaults".
+const addAlert = (alert: object) => {
+	event.followsDefaults = false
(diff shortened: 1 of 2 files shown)


krantheman:
@greptileai review

Re the open threads:

- **Default state overrides removal** — stale as of 3085b3b: every add now goes through `addAlert`, which clears `followsDefaults`, so add-then-remove really does send the clear.
- **Value matching moves authored reminders** — intentional. After gating on all-day, the only "authored" alert this can touch is one hand-edited to be byte-identical to the default (9am on the start date). It's indistinguishable by value, and following the event's date move is the behaviour you'd want there anyway; provenance tracking isn't worth the extra state for that case.

PR #792, opened 2026-09-01
fix(calendar): organize new events as the default participant identity

s-aga-r:
## Summary
- The event modal seeded the organizer and the pre-filled self-participant row from the Frappe login user (`user.data.name`), which can differ from the account's Stalwart identities. When it did, `acting_as_organizer()` returned false on create and custom invite emails were silently skipped in favour of server scheduling.
- The calendar user store now fetches the account's mail Identities alongside its Participant Identities and exposes `organizerIdentity`: among the addresses present in both lists (invites go out as mail from the organizer, so it must be able to send), the participant identity flagged `default` wins, else the first. `undefined` until both lists have loaded, or when no address is on both. New events take their organizer and self-participant row from it; editing an existing event still carries over its stored organizer.
- Handled on the UI side: while a new event has no organizer, the Save split stays disabled with a tooltip ("No identity to organize the event. Add a participant identity in Calendar Settings whose email is also one of the account's mail identities.") and closing a dirty modal offers Discard instead of silently saving a draft. If the identity lists land after the modal opened, the organizer row is seeded then.
- `create_scheduled_meeting` (the "Add Meet link" path from the same modal) no longer overrides the organizer with the JMAP Account's `default_outgoing_email` / session user. It passes the client's organizer through like `add_calendar_event`; an omitted organizer resolves to the default participant identity in the JMAP layer.
- Vitest coverage for `organizerIdentity`: unloaded lists, no common address, flagged default that can send, a default that cannot send being skipped, first-common fallback.

## Test plan
- `cd frontend && npx vitest run src/apps/calendar`: 4 files, 39 tests pass.
- `ruff check` / `ruff format --check` on `suite/meet/api/schedule.py`: clean.
- Manual: on an account whose login email differs from its default Participant Identity (with a matching mail Identity), create an event with another attendee from the Calendar UI; the organizer badge shows the identity's address and the invite email goes out. Repeat with "Add Meet link" enabled. Then make the default Participant Identity's email one with no mail Identity: the first participant identity that does have one organizes instead. Finally remove every matching pair and open "Add Event": Save and "Save as draft" are disabled, hovering the split shows the tooltip, and closing a dirty modal asks to discard rather than saving a draft.

### DIFF
--- a/.github/scripts/coverage_report.py
+++ b/.github/scripts/coverage_report.py
@@ -58,7 +58,7 @@ def delta(current: float | None, baseline: float | None) -> str:
         return ""
     change = current - baseline
     if abs(change) < 0.05:
-        return " (no change)"
+        return " (+0%)"
     return f" ({change:+.1f}%)"
 
 
--- a/.github/scripts/test_coverage_reporting.py
+++ b/.github/scripts/test_coverage_reporting.py
@@ -55,6 +55,9 @@ class CoverageReportTest(unittest.TestCase):
         self.assertIn("Backend 85.0% (+5.0%)", markdown)
         self.assertIn("**70.0%** (70 / 100) (-2.0%)", markdown)
 
+    def test_renders_rounded_zero_delta_as_percentage(self):
+        self.assertEqual(coverage_report.delta(34.6, 34.6), " (+0%)")
+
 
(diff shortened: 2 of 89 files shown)


s-aga-r:
@greptileai review

PR #809, opened 2026-09-02
feat(calendar): ask how far a recurring change reaches

krantheman:
Recurring events asked "this instance or the whole series?" with a button per answer. The far-reaching answer carried the primary style, there was no Cancel besides the ✕, and two buttons couldn't grow a third answer — which delete had needed all along and had solved on its own with a submenu: three commands behind a hover, each firing the moment it was touched.

One dialog now, asked the same way everywhere it comes up — the edit modal, a dragged occurrence, delete, and RSVP, which never asked at all. The answers are a radio list, the safest is pre-selected so Enter does the least far-reaching thing, and a single primary button carries the verb.

All three answers work in all four places. "This and following" is dropped at the head of a series, where it means the same as "all events".

## Making the answers real

Only "all events" and (for delete) "this instance" existed before. The rest is new:

- **Edit one occurrence** — writes an override addressed by `(master_id, recurrence_id)`, wired in the edit modal and for a dragged occurrence, which used to undo the move rather than leave it on screen as if saved.
- **This and following** — JSCalendar can say "this date" or "the series" and nothing between, so the series is cut in two: the original stops before this occurrence, the edit becomes a new series starting at it. A counted series is shared out rather than repeated, and overrides the truncated rule no longer generates move to the new half.
- **Delete this and following** — the same truncation, server-side now. Three bugs went with the move: `until` was set to midnight on the occurrence's date (so an all-day series kept the day it was told to delete), a counted series got an `until` bolted on beside its `count`, and it hung on a field the event doesn't always carry — writing `undefinedT00:00:00Z` when absent.
- **Answer one occurrence** — the status lands as an override on that date and the iTIP reply names it with a `RECURRENCE-ID`.

## Fixes found while testing it

Most of the diff. Nearly all are one root cause: an override losing track of the thing that generates it.

- Moving a series left its overrides on their old keys, so an edited occurrence was drawn twice — once where the rule now puts it, once where it used to be. Keys shift with the series. Same for a rule whose day selectors were reanchored by the move, which reads differently while describing the same series.
- A rule names its days by reading them off the start and doesn't re-read them when the anchor moves. A Monday series moved to a Wednesday started on Wednesday and repeated on Mondays, so the occurrence just moved wasn't drawn at all.
- Dragging a series by its second occurrence sent that occurrence's start as the master's, re-basing every date after it — the first occurrence disappeared.
- Stalwart 0.16.20 returns an occurrence carrying an override without its title, description, privacy, locations, participants or organizer, and with a duration invented to run to midnight. The series fills those back in, using its own `recurrenceOverrides` to know which properties the occurrence actually owns.
- An invitation is delivered as an occurrence while the account keeps its own copy of the same uid; answering filled that copy in and drew it as a second, untitled event.
- An override doesn't replace the occurrence it names — both come back. The one to draw is the one synthesised from the override, which the server never stored and so has no creation time.
- An occurrence's link used the row's id, which is synthetic and moves when that occurrence gains an override. The panel lost its event on the first edit.
- `getRepeatMessage` asserted its way to a frequency, so an occurrence whose series lost its rule took the whole detail panel down.

## Testing

Frontend unit tests for the option lists, the reanchor and the repeat message; Python unit tests for the split arithmetic.

The rest is verified by hand against Stalwart 0.16.20, including a real invitation sent between two accounts — edit/delete/move at each scope, the split's count sharing, overrides surviving a move and a truncation, an invitation drawn once, and per-occurrence and series-wide answers alongside each other. There is no automated integration coverage: `StalwartIntegrationTestCase` can't create a member on 0.16.20 (`locale: "en_US"` is rejected, and admin impersonation then returns 401), so CI still pins 0.16.17 and these paths only run locally.


krantheman:
@greptileai review

PR #811, opened 2026-09-02
fix(mail): delete the site's push subscriptions when a user is disabled

s-aga-r:
## Summary
- Disabling a user now deletes this site's push subscriptions on the mail server. Only subscriptions carrying the site's deterministic device client id are removed; the user's custom subscriptions stay.
- New `delete_site_push_subscriptions` helper in the Push Subscription module, sharing the per-id error reporting with the existing delete path.
- New `delete_push_subscriptions_on_disable` User handler, first in the `on_update` list so it runs ahead of `apply_disabled_account_role`, whose Stalwart role may cut the JMAP access the deletion needs.
- The JMAP connection factory gains an explicit `allow_disabled` keyword, passed only by this cleanup, so it can reach the mail server for a user whose record already reads as disabled. Requests never pass it.
- Re-enable needs no counterpart: login healing recreates the site subscription on the user's next sign-in.

## Test plan
- 10 new unit tests in `suite/mail/tests/test_push_subscription_healing.py` cover filtering by device client id, skipping the delete when nothing matches, surfacing server-side errors, the handler's guards, log-not-raise on failure, and the connection factory refusing disabled users unless `allow_disabled` is passed while still refusing missing users. Module passes (37 tests).
- Verified live on a dev site by disabling a JMAP-configured user inside a rolled-back transaction with the User document cache cleared: the handler ran before the disabled-role handler, the connection opened with the record reading disabled, the delete succeeded, and no Error Log entry was written.

### DIFF
--- a/suite/hooks.py
+++ b/suite/hooks.py
@@ -253,12 +253,10 @@ doc_events = {
             "suite.drive.utils.users.create_drive_settings",
             "suite.mail.events.create_user_settings",
         ],
-        # Before save, not on update: the mail server is still reachable for the user while
-        # the record reads as enabled (see delete_push_subscriptions_on_disable).
-        "before_save": [
-            "suite.mail.events.delete_push_subscriptions_on_disable",
-        ],
         "on_update": [
+            # First: the disabled account role applied below may cut the JMAP access the
+            # deletion needs.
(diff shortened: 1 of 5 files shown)


s-aga-r:
@greptileai review

PR #830, opened 2026-09-07
fix(mail): check the account belongs to the caller before reading it

krantheman:
`get_mailboxes` and `get_quota` took an account id from the caller and checked only that the caller had a JMAP account of their own — never that this one was it. Both read the local tables through `frappe.get_all`, which bypasses permissions, and neither reaches a JMAP service, so there was no ownership check further down to fall back on either.

Both now call `is_jmap_account_belongs_to_user`, as the six other account-scoped endpoints in `mail.py` already did. Tests cover both endpoints in both directions — the owner reads their own, another user is refused.

`get_quota` wasn't in the report; it turned up looking for the same shape elsewhere. Two other candidates were false alarms and are left alone: `verify_otp` takes a signup token rather than an account, and `get_address_book_contact_count` resolves ownership through its JMAP service.


Darshan060224:
Could you also let me know whether a security advisory will be published for this issue, and whether a CVE will be assigned? I’d also appreciate being credited as the reporter in the advisory,

PR #851, opened 2026-09-11
feat(calendar): name the mailing list in the To header of member invites

s-aga-r:
## Summary
- A mailing list invited to an event now stays on it as a group participant with `scheduleAgent: none`, and each member added for it points back at the list through `memberOf` (RFC 8984 §4.4.6). Stalwart round-trips both fields.
- Invitation, update and cancellation mail to a member now carries the list in its **To** header, the way any other mail to the list arrives. The SMTP envelope, the `.ics` ATTENDEE and the RSVP link stay the member's own, so each member still gets their own copy and their own response link.
- The list itself is never mailed: the sender skips unscheduled participants, and the update diff ignores them so removing nothing never produces a stray cancel.
- The iTIP ATTENDEE records the membership with `MEMBER="mailto:list@..."`, and the group line goes out with `RSVP=FALSE`.
- Re-expansion on every save is now membership-aware: members still on the list keep their entry and recorded RSVP, members who left are dropped (and get a cancellation), and removing the list from the event uninvites its members. A list an admin deleted from the directory leaves its members alone.
- The size cap counts invited participants only; the display-only list entry does not use a slot.
- `Event Participant` gains `schedule_agent` and `member_of`; the JMAP serialiser and read model carry both. No frontend change: it sends the API's participant objects back unchanged.

Only the custom-invite path is affected. When custom invites are off, Stalwart builds the iMIP itself and keeps naming the member.

## Test plan
- [x] `bench --site s2 run-tests --module suite.calendar.tests.test_calendar_mailing_list_participants` passes (29 tests), including new cases for later saves, list removal, directory deletion, To-header resolution and the ATTENDEE output.
- [x] End-to-end against a live JMAP server: stored list-shaped participants through `add_calendar_event`, read them back, and ran `notify_participants` with the queue write captured. The member's mail read `To: Team <team@example.com>` with only the member in the envelope; the list got no mail; the ATTENDEE line carried `MEMBER`.
- [x] `TestMailingListInvite` (real Stalwart directory) needs a site with Stalwart admin credentials; it skips on the local bench.


s-aga-r:
@greptileai review

s-aga-r:
@greptileai review

PR #903, opened 2026-09-19
feat(mail): delete identities from settings

s-aga-r:
## Summary
- Mail UI > Settings > Identity gains a Delete button with a confirm dialog, mirroring Participant Identity in Calendar settings. Until now identities could only be deleted from the desk.
- The button is shown only when the server reports the identity as deletable (JMAP `mayDelete`, exposed as `may_delete`). Deletion goes through the existing whitelisted `Identity.bulk_delete`.
- The selection stays valid after create/delete/account switch by falling back to the first identity, and the page shows an empty state once the last identity is gone (Stalwart does not re-create the primary one).
- The Stalwart integration test for identity settings now deletes through `bulk_delete` by document name, the path the UI uses, and asserts the `may_delete` flag the button keys on.

## Test plan
- Verified on a live Stalwart-backed account: created a throwaway identity, deleted it via the new button and confirm dialog, confirmed via the API it was gone and the selection fell back to the remaining identity.
- Deleted the account's only identity: empty state appears, Delete button disappears; re-created it via New and the form returned.
- `cd frontend && npx vitest run src/apps/mail` passes (277 tests). `ruff` passes on the test file.
- `suite.mail.tests.test_mail_identity_settings` needs a Stalwart admin credential to run; not executed locally.


s-aga-r:
@greptileai review

